PR #7, opened 2019-12-20
Feature/core 7828 wizard layout

riboher:
Computed properties for slots are not reactive. I removed the computed property that checked the presence of slots and added directly to the template.

### DIFF
--- a/src/elements/GeoWizardLayout/GeoWizardLayout.vue
+++ b/src/elements/GeoWizardLayout/GeoWizardLayout.vue
@@ -21,7 +21,7 @@
       </section>
 
       <footer
-        v-if="!!this.$slots.footerTrailing || !!this.$slots.footerLeading"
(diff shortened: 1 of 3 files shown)


src/elements/GeoWizardLayout/GeoWizardLayout.vue
```diff
@@ -21,7 +21,7 @@
       </section>
 
       <footer
-        v-if="hasFooter"
+        v-if="!!this.$slots.footerTrailing || !!this.$slots.footerLeading"
```
dpramar:
```suggestion
        v-if="!!$slots.footerTrailing || !!$slots.footerLeading"
```
Should be without `this`, right?

riboher: True! 🤦‍♂ Fixed now!

PR #11, opened 2019-12-31
fix GeoInput width and days in GeoCalendar

soleneG02:


src/styles/elements/GeoInput/_geo-input.scss
```diff
@@ -93,6 +93,11 @@
     color: $color-red;
     cursor: pointer;
   }
+  
+  .geo-input__icon--hidden {
```
fofi: What does "hidden" stands for here? why the color is white?

src/styles/elements/GeoInput/_geo-input.scss
```diff
@@ -93,6 +93,11 @@
...
+    color: $color-white;
+    cursor: text;
```
fofi: Why cursor text?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 29.1.2
```
riboher: Please change this to 29.1.3. I have a previous PR with this version. Also, some tests in GeoCalendar are failing (I guess because of the new year), so I'd try to fix them before merging these

src/elements/GeoInput/GeoInput.vue
```diff
@@ -42,9 +42,12 @@
           <!-- blur event won't be fired but that's fine because we want this handler to prevail over the blur one -->
           <!-- https://forum.vuejs.org/t/blur-before-click-only-on-safari/21598/7 -->
           <font-awesome-icon
-            v-if="isDeleteIconVisible"
+            v-if="isDeleteIconPresent"
```
riboher: Why the change from visible to present? We should stick to one naming

src/elements/GeoInput/GeoInput.vue
```diff
@@ -161,8 +164,20 @@ export default {
       return null
     },
 
+    isDeleteIconPresent () {
```
riboher: Let's call this property `hasDeleteIcon`

src/elements/GeoInput/GeoInput.vue
```diff
@@ -42,9 +42,12 @@
           <!-- blur event won't be fired but that's fine because we want this handler to prevail over the blur one -->
           <!-- https://forum.vuejs.org/t/blur-before-click-only-on-safari/21598/7 -->
           <font-awesome-icon
-            v-if="isDeleteIconVisible"
+            v-if="isDeleteIconPresent"
             :icon="deleteInputValueIcon"
-            class="geo-input__icon--delete geo-input__accessory-items-item"
+            :class="{
+              [deleteIconClass]: true,
```
riboher: I'm not sure I like this, I would be more explicit and do something like
`:class="{
   'geo-input__icon--delete': isDeleteIconVisible,
   'geo-input__icon--hidden': !isDeleteIconVisible
  }"`

It's a little bit more verbose but it gives more information. But again, just an opinion

src/elements/GeoInput/GeoInput.vue
```diff
@@ -161,8 +164,20 @@ export default {
...
-      return !this.disabled && !!this.value && !!this.$listeners['delete-value']
+      return this.isDeleteIconPresent && !!this.value
+    },
+
+    deleteIconClass () {
+      if (this.isDeleteIconVisible) {
+        return 'geo-input__icon--delete'
+      } else {
+        return 'geo-input__icon--hidden'
+      }
```
riboher: If we're going to leave this, let's do something like
```
return this.isDeleteIconVisible
  ? 'geo-input__icon--delete'
  : 'geo-input__icon--hidden'
```

src/elements/GeoInput/GeoInput.vue
```diff
@@ -161,8 +165,18 @@ export default {
...
+    hasDeleteIcon () {
+      return !this.disabled && !!this.$listeners['delete-value']
+    },
+
     isDeleteIconVisible () {
-      return !this.disabled && !!this.value && !!this.$listeners['delete-value']
+      return this.hasDeleteIcon && !!this.value
+    },
+
+    deleteIconClass () {
```
fofi: It looks like we are not using this property

src/elements/GeoInput/GeoInput.vue
```diff
@@ -42,9 +42,13 @@
           <!-- blur event won't be fired but that's fine because we want this handler to prevail over the blur one -->
           <!-- https://forum.vuejs.org/t/blur-before-click-only-on-safari/21598/7 -->
           <font-awesome-icon
-            v-if="isDeleteIconVisible"
+            v-if="hasDeleteIcon"
             :icon="deleteInputValueIcon"
-            class="geo-input__icon--delete geo-input__accessory-items-item"
+            :class="{
+              'geo-input__icon--delete': hasDeleteIcon,
+              'geo-input__icon--hidden': !isDeleteIconVisible,
```
fofi: It looks like now is possible to have both classes, do we want that?

riboher: As I understand, the icon can be both visible and present. Present doesn't mean that it's visible, but if it's visible, it's definitely present

test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
```diff
@@ -23,7 +23,7 @@ import getYear from 'date-fns/getYear'
 import startOfYear from 'date-fns/startOfYear'
 import subDays from 'date-fns/subDays'
 
-const today = new Date(2019, 6, 30) // Fixed date to avoid future errors with random dates
```
riboher: Can you create a task to solve this in a more permanent way? I thought this would prevent these kind of errors and it didn't, and we can't keep changing these when the year changes. For now it's ok though

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 29.1.3
```
dpramar: You shouldn't do the versioning taking into account other pull requests, otherwise this is blocked until the other is merged but there is no need.

src/elements/GeoInput/GeoInput.vue
```diff
@@ -161,8 +164,18 @@ export default {
...
+    },
+
     isDeleteIconVisible () {
-      return !this.disabled && !!this.value && !!this.$listeners['delete-value']
+      return this.hasDeleteIcon && !!this.value
+    },
+
+    deleteIconClass () {
+      return this.isDeleteIconVisible
+        ? 'geo-input__icon--delete'
```
dpramar: I would name this class `geo-input__delete-icon--visible` and `geo-input__delete-icon--hidden`, in BEM after the `--` should be a modifier.

src/styles/elements/GeoInput/_geo-input.scss
dpramar: Probably this class should just have `visibility: hidden;`

test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
dpramar: This might be fixed if we set an `initialDateInGrid` in `getWrappedComponent` so the calendar is opened at the same day and year always.

test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
```diff
@@ -23,7 +23,7 @@ import getYear from 'date-fns/getYear'
...
+const today = new Date(2020, 6, 30) // Fixed date to avoid future errors with random dates
```
riboher: This looks like we'll be needing to keep changing this year after year 😅. Is this the issue you're addressing in the other PR?

PR #12, opened 2020-01-07
fix GeoCalendar tests

soleneG02:
- set `currentYear` to a fixed one before selecting month or quarters
- use fixed Date for `GeoCalendarNavigationDay` and `GeoCalendarNavigationMonth`

### DIFF
--- a/test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
+++ b/test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
@@ -839,19 +839,6 @@ function getWrappedComponent () {
       latestDatePlaceholder: 'Latest date',
       pickerGranularity: '<div> Granularity Selectors </div>'
     },
-    data () {
-      return {
-        fromFormattedDate: '',
-        toFormattedDate: '',
-        fromRawDate: null,
-        toRawDate: null,
-        currentMonth: 6,
-        showFromFormatError: false,
-        showToFormatError: false,
-        currentInitialYearInRange: 0,
-        currentEndYearInRange: 0
-      }
-    },
     propsData: {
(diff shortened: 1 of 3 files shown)


test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
```diff
@@ -835,7 +846,6 @@ function getWrappedComponent () {
         fromRawDate: null,
         toRawDate: null,
         currentMonth: 6,
-        currentYear: 2019,
```
dpramar: This means that this whole object is not used at all?
In the vue test utils docs I don't see that `mount` accepts a `data` function, can you delete `data ()` if it's not used?

test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
```diff
@@ -229,15 +238,17 @@ describe('GeoCalendar', () => {
...
         wrapper.vm.selectMonth(initialMonth)
         wrapper.vm.selectMonth(getMonth(endDate))
         geoFromInput.vm.$emit('focus')
-
         wrapper.setData({
           currentYear: getYear(invalidFromDateRange)
         })
 
-        wrapper.vm.selectMonth(getMonth(invalidFromDateRange))
+        wrapper.vm.selectMonth(getMonth(invalidFromDateRange), true)
```
dpramar: What means the `true`?

test/unit/specs/elements/GeoCalendar/GeoCalendar.spec.js
soleneG02: Not used indeed, I deleted it

PR #15, opened 2020-01-09
add tooltip props in GeoTrimmedContent

soleneG02:


src/elements/GeoTrimmedContent/GeoTrimmedContent.vue
```diff
@@ -32,7 +34,29 @@ export default {
...
   status: 'ready',
   release: '10.1.0',
-  directives: { OnResize, Tooltip },
+  directives: { OnResize },
+  props: {
+    tooltipPosition: {
+      type: String,
+      default: POSITIONS.top,
+      validator (value) {
+        if (value in POSITIONS) return true
```
fofi: Be aware that here we are checking that **value** is a possible key in POSITIONS, not a possible value inside the POSITIONS object

src/elements/GeoTrimmedContent/GeoTrimmedContent.vue
```diff
@@ -32,7 +34,29 @@ export default {
...
+
+        const supportedValues = Object.values(POSITIONS).map(i => `«${i}»`).join(', ')
+        console.warn(`GeoTrimmedContent [component] :: Unsupported value («${value}») for «tooltipPosition» property. Use one of ${supportedValues}`)
+      }
+    },
+    tooltipAlignment: {
+      type: String,
+      default: ALIGNMENTS.middle,
+      validator (value) {
+        if (value in ALIGNMENTS) return true
```
fofi: same here as before

src/elements/GeoTrimmedContent/GeoTrimmedContent.vue
fofi: now is working because the key and value are the same, but not sure if we should asume that

CHANGELOG.md
```diff
@@ -1,11 +1,18 @@
...
 
 - `GeoTooltip` position and alignment can be passed to `GeoTrimmedContent`
 
+## 29.1.3
+
+Fix :
+
+- `GeoInput` width doesn't change when delete icon appears
+- `GeoCalendar` day icons are correctly displayed when creating an interval
+
```
fofi: these changes are from 29.1.4, right?

soleneG02: From 29.1.3, but that's another PR I just merged
By the way, this PR should be 29.2.0 and not 29.1.4, I will change it

src/elements/GeoTrimmedContent/GeoTrimmedContent.vue
soleneG02: That's how it's done in the other components of the DS, and the goal is to "force" the use of keys instead of direct string so I don't think it's a problem

fofi: But then, why the `supportedValues` are the values inside POSITIONS?

soleneG02: You're right, I changed it in `enumPropertyFactory.js` file, to test the values and not the keys
Should I change it also for the other times where we use this kind of validator ?

fofi: if there are a lot of them, lets create a task to do it, if not just do it here

soleneG02: There are at least 34 so I'll create a new task!

src/elements/GeoTooltip/GeoTooltip.vue
```diff
@@ -31,14 +31,14 @@ const getNextStaticTooltipId = counterFactory()
 /** @type {Element|null} */
 let tooltipContainerElement = null
 
-const POSITIONS = {
+export const POSITIONS = {
```
dpramar: Better move this constants to a new file, like `GeoTooltip.constants.js` and export them there.

Then import in this component and in `GeoTrimmedContent`.

It's just because importing in `GeoTrimmedContent` the component GeoTooltip.vue for the constants looks strange.

src/utils/enumPropertyFactory.js
```diff
@@ -10,9 +10,10 @@ export default function enumPropertyFactory (params) {
     type: String,
     default: params.defaultValue,
     validator (value) {
-      if (value in params.enumDictionary) return true
+      const allValues = Object.values(params.enumDictionary)
+      if (allValues.includes(params.enumDictionary[value])) return true
```
dpramar: Shouldn't this be 

```suggestion
      if (allValues.includes(value)) return true
```

src/elements/GeoTooltip/GeoTooltip.constants.js
```diff
@@ -0,0 +1,17 @@
+const POSITIONS = {
+  keyyy: 'valuuue',
```
riboher: This doesn't seem like a correct position 😅

soleneG02: Yes I changed it to make tests and forgot to put it back..

PR #17, opened 2020-01-10
use enumPropertyFactory validator to check values

soleneG02:


src/elements/GeoDropdown/GeoDropdown.vue
```diff
@@ -117,9 +105,10 @@ export default {
...
       validator: function (value) {
-        if (value === undefined || value in GeoDropdownConstants.Y_AXIS_POSITION) return true
+        const allValues = Object.values(GeoDropdownConstants.Y_AXIS_POSITION)
+        if (value === undefined || allValues.includes(value)) return true
 
-        const supportedValues = Object.values(GeoDropdownConstants.Y_AXIS_POSITION).map(i => `«${i}»`).join(', ')
+        const supportedValues = allValues.map(i => `«${i}»`).join(', ')
         console.warn(`GeoDropdown [component] :: Unsupported value («${value}») for «forceYAxisPosition» property. Use one of ${supportedValues}`)
         return false
       }
```
riboher: Why not using the enumPropertyFactory here?

src/elements/GeoInput/GeoInputMessage.vue
```diff
@@ -43,10 +43,10 @@ export default {
...
-        if (!value) return true
-        if (value in VARIANTS) return true
+        const allValues = Object.values(VARIANTS)
+        if (!value || allValues.includes(value)) return true
 
-        const supportedValues = Object.values(VARIANTS).map(i => `«${i}»`).join(', ')
+        const supportedValues = allValues.map(i => `«${i}»`).join(', ')
         console.warn(`GeoInputMessage [component] :: Unsupported value («${value}») for «variant» property. Use one of ${supportedValues}`)
         return false
       }
```
riboher: Same, can't you use it here as well?

src/elements/GeoSelect/GeoSelect.mixin.js
```diff
@@ -208,7 +209,12 @@ export default {
...
       required: false,
       validator: function (value) {
-        return value === undefined || value in Y_AXIS_POSITION
+        const allValues = Object.values(Y_AXIS_POSITION)
+        if (value === undefined || allValues.includes(value)) return true
+
+        const supportedValues = allValues.map(i => `«${i}»`).join(', ')
+        console.warn(`GeoSelect [component] :: Unsupported value («${value}») for «forceYAxisPosition» property. Use one of ${supportedValues}`)
+        return false
       }
```
riboher: Same. Use it here?

src/elements/GeoDropdown/GeoDropdown.vue
fofi: I think because here we are allowing undefined

src/utils/enumPropertyFactory.js
```diff
@@ -1,18 +1,46 @@
...
+ * @param {object} params
+ * @param {Boolean} [params.required]
+ * @param {Object<string, string>} params.enumDictionary
+ * @param {string} params.componentName
+ * @param {string} params.propertyName
+ */
+function enumRequiredPropertyFactory (params) {
+  return {
+    type: String,
+    required: params.required ? params.required : false,
```
fofi: We can simple cast the value to boolean using `!!params.required`

src/utils/enumPropertyFactory.js
```diff
@@ -1,18 +1,46 @@
...
+
 /**
  * @param {object} params
  * @param {string} [params.defaultValue]
  * @param {Object<string, string>} params.enumDictionary
  * @param {string} params.componentName
  * @param {string} params.propertyName
  */
-export default function enumPropertyFactory (params) {
+function enumDefaultPropertyFactory (params) {
```
fofi: Maybe we can use the same function, and depending on the parameter add one property or other.
For example
```
const defaultEnumProperty = {
  type: String,
  validator
  ....
}
if ('defaultValue' in params) {
  defaultEnumProperty.default = params.defaultValue
}
if (required...
```

fofi: And maybe we could also add a property in params to check `value === undefined`.

We can discuss it later on which approach take

fofi:
Nice work on the refactor :recycle:

src/elements/GeoInput/GeoInputMessage.vue
soleneG02: It was because of the !value as well, but I introduced it in the function

PR #19, opened 2020-01-14
Align texts when list item has icon

dpramar:


src/elements/GeoList/GeoListItem.vue
```diff
@@ -37,7 +37,10 @@
     </div>
     <div
       v-if="$slots.description"
-      class="geo-list-item__description"
+      :class="{
+        'geo-list-item__description': true,
+        'geo-list-item__description--spaced': props.icon
```
riboher: Is it better to cast it as a boolean? Seems the correct thing to do if icon is an array to do `!!props.icon`

PR #20, opened 2020-01-20
Feature/core 7965 geo input fixed width dropdown

riboher:
The main issue to solve here was the fact that if we wanted to implement an editable input with dynamic width (this being the input grow and shrink depending on the content), the buttons of the dropdown remained in the same place, not updating its position when the width of its parent container changed.

This introduces a new prop (repositionOnUpdate) that forces a popup reposition when the component updates.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 30.0.0
```
fofi: Why is this a breaking change? aren't we just adding an optional property?

src/elements/GeoDropdown/GeoDropdown.vue
```diff
@@ -124,6 +124,11 @@ export default {
     popupClass: {
       type: [String, Array, Object],
       required: false
+    },
+
+    repositionOnUpdate: {
```
fofi: Missing comment about this property

CHANGELOG.md
riboher: Hmmm, you're probably right. Since it was going to affect somehow something in the app, I thought it should be a major. I'll change it

PR #21, opened 2020-01-21
GeoChartStackedBars component

soleneG02:


src/elements/GeoChart/GeoChart.11.stacked-bars.examples.md
```diff
@@ -0,0 +1,366 @@
+## Stacked bar charts
+
+Stacked bar charts are collections of grouped items within a bar chart item which are displayed as rectangles in a 2-dimensional grid. An arbitrary amount of different collections of items can
+be displayed using statcked bar charts, each of those collections are called **groups**.
```
riboher: Typo: using `stacked` bar charts

src/elements/GeoChart/GeoChart.11.stacked-bars.examples.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Required properties
+
+Each **group** **requires** these properties:
```
riboher: You don't need to treat each word as independent bold words. You can just do `**group requires**`

src/elements/GeoChart/GeoChart.11.stacked-bars.examples.md
```diff
@@ -0,0 +1,366 @@
...
+corresponding to the stacked bar being customized and its position inside the data array.
+It's expected to return a HTML string that will be rendered inside a tooltip.
+- `offset` - *Optional*. Function that takes as parameter the event triggering the
+tooltip and is expected to return an object with an `x` and a `y` property, both
+storing numbers that will be used as offset of the tooltip with respect to event
+coordinates. By default tooltip will be positioned above cursor.
+
+### Customizing CSS classes
+
+Each stcaked bar can customize its CSS classes by setting a function for key `cssClasses`.
```
riboher: Typo: each `stacked` bar

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+    .data(d => d.data)
+
+  barWrapper
+    .attr('height', getBarInitialHeight)
+    .attr('width', getBarInitialWidth)
+
+  const newBarWrapper = barWrapper
+    .enter()
+    .append('g')
+    .attr('class', 'geo-chart-stacked-bars-group__single-group')
```
riboher: Could we need also to insert the index in the class as we're doing in the render function?

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+    ? dimensionUtils.DIMENSIONS_2D.vertical
+    : dimensionUtils.DIMENSIONS_2D.horizontal
+
+  const barWrapper = group
+    .selectAll('g.geo-chart-stacked-bars-group__single-group')
+    .data(d => d.data)
+
+  barWrapper
+    .attr('height', getBarInitialHeight)
+    .attr('width', getBarInitialWidth)
```
riboher: Do you actually need this chunk? You're already setting the initial height and width on the enter part on lines 102-107

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+      },
+      getTranslationForNormalAxis: axisUtils.getTranslationForNormalAxisFactoryStackedBar
+    })(d, i)
+  }
+
+  function getBarTransform (d, i) {
+    const barNormalTranslation = getTranslation(d, i)
+    const widthTranslation = isMainDimensionHorizontal
+      ? -getWidthTranslation(d, i) / 2
+      : -getWidthTranslation(d, i) / 2
```
riboher: This can't be right, can it? You're assigning the same thing no matter the condition. If it doesn't matter the dimension, then this should be
`const widthTranslation = -getWidthTranslation(d, i) / 2`

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+    .attr('height', getBarHeight)
+    .attr('width', getBarWidth)
+
+  barWrapper
+    .exit()
+    .remove()
+
+  allBarWrappers.each(function (barData) {
+    const singleGroupBar = d3.select(this)
+    renderStackedBars(singleGroupBar, barData, singleGroupOptions, globalOptions, {
```
riboher: Let's call barData `stackedBarData` instead

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+ * @param {d3.Selection<GElement, Datum, PElement, PDatum>} stackedBarsContainer
+ * @param {Object} barData
+ * @param {GeoChart.SingleStackedBarGroupConfig<HorizontalDomain, VerticalDomain>} singleGroupOptions
+ * @param {GeoChart.GlobalOptions} globalOptions
+ * @param {Object} params
+ * @param {GeoChart.AxisConfig<MainDimensionDomain, any>} params.axisForMainDimension
+ * @param {GeoChart.AxisConfig<NormalDimensionDomain, any>} params.axisForNormalDimension
+ * @param {d3.Tooltip<SVGElement, object, PElement, PDatum>} [d3TipInstance]
+ */
+function renderStackedBars (stackedBarsContainer, barData, singleGroupOptions, globalOptions, {
```
riboher: Same here, I think this should be called `stackedBarData`

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+
+  let temporaryBarWrapperHeight = 0
+  const allStackedBarsSegments = _.times(stackedBarsSegmentsData.length, function (idx) {
+    const stackedBarSegment = {
+      startValue: idx === 0
+        ? Math.min.apply(Math, axisForMainDimension.scale.axisScale.domain())
+        : temporaryBarWrapperHeight,
+      endValue: idx === stackedBarsSegmentsData.length - 1
+        ? Math.max.apply(Math, axisForMainDimension.scale.axisScale.domain())
+        : temporaryBarWrapperHeight + stackedBarsSegmentsData[idx][axisForMainDimension.keyForValues]
```
riboher: As we've discussed, this can probably be easily solved with lodash `_.min(axisForMainDimension.scale.axisScale.domain())`

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+    }
+    temporaryBarWrapperHeight += stackedBarsSegmentsData[idx][axisForMainDimension.keyForValues]
+    return stackedBarSegment
+  })
+
+  const stackedBarsOriginAtAxis = axisForMainDimension.scale.axisScale(axisForMainDimension.scale.valueForOrigin)
+
+  const directionToChange = isMainDimensionHorizontal
+    ? 'x'
+    : 'y'
```
riboher: I think this could be done differently doing an object instead of having to check the condition each time. Something maybe like 
`
const axisPositions = {
  [singleGroupOptions.axis.horizontal]: 'x',
  [singleGroupOptions.axis.vertical]: 'y',
}
const directionToChange = axisPositions[singleGroupOptions.mainDimension]
`

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,367 @@
...
+
+  const stackedBars = stackedBarsContainer
+    .selectAll(`rect.${stackedBarsBaseClass}`)
+    .data(allStackedBarsSegments)
+
+  stackedBars
+    .attr(directionToChange, stackedBarsOriginAtAxis)
+    .attr('width', getStackedBarsInitialWidth)
+    .attr('height', getStackedBarsInitialHeight)
```
riboher: Same as before. Do you actually need this chunk if you're setting all of the properties below?

soleneG02: In this case it's needed because even the bars that aren't new are supposed to make a transition from 0 to their height when we change the graph

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,365 @@
...
+  axisForNormalDimension
+}, d3TipInstance) {
+  const isMainDimensionHorizontal = axisUtils.isMainDimensionAxis(singleGroupOptions.axis.horizontal, singleGroupOptions)
+
+  const stackedBarsBaseClass = 'geo-chart-stacked-bars__segment'
+
+  const stackedBarsSegmentsData = stackedBarData[axisForMainDimension.keyForValues]
+
+  let temporaryBarWrapperHeight = 0
+  const allStackedBarsSegments = _.times(stackedBarsSegmentsData.length, function (idx) {
```
dpramar: This is "modifying" what the user send as data. We don't know how each object of `stackedBarsSegmentsData` is, so we should load the segments as they come or add new properties like `startValue` and `endValue`, but never delete or modify.

As example, now in the tooltip as `d` you only have an object with `startValue` and `endValue`.

Could this be done iterating `stackedBars`? something like this: (might be missing something)

```
stackedBars
    .each(function (d, i) {
        d.startValue = i === 0
            ? _.min(axisForMainDimension.scale.axisScale.domain())
            : temporaryBarWrapperHeight
        d.endValue = i === stackedBarsSegmentsData.length - 1
            ? _.max(axisForMainDimension.scale.axisScale.domain())
            : temporaryBarWrapperHeight + d[axisForMainDimension.keyForValues]
        temporaryBarWrapperHeight += d[axisForMainDimension.keyForValues]
    })
```

src/elements/GeoChart/GeoChartUtils/axisUtils.js
```diff
@@ -87,6 +87,12 @@ export function getTranslationForNormalAxisFactory (options, { keyForNormalOffse
   }
 }
 
+export function getTranslationForNormalAxisFactoryStackedBar (normalAxis, singleItem) {
+  const positionOfItemValue = axisUtils.getItemValueAtAxis(normalAxis, singleItem)
+  const normalOffset = normalAxis.scale.axisScale.bandwidth() / 2
```
fofi: Why not doing the same as we are doing for the generic function above (getTranslationForNormalAxisFactory)
```
    const naturalNormalOffset = numericUtils.isNumberForced(options, keyForNaturalNormalOffset)
      ? options.naturalNormalOffset	      ? options.naturalNormalOffset
      : 0	      : 0
```

src/elements/GeoChart/GeoChartUtils/axisUtils.js
```diff
@@ -87,6 +87,12 @@ export function getTranslationForNormalAxisFactory (options, { keyForNormalOffse
   }
 }
 
+export function getTranslationForNormalAxisFactoryStackedBar (normalAxis, singleItem) {
```
fofi: This is not really a factory, as we are actually returning the translation function, look for example at getTranslationForNormalAxisFactory, there we are actually creating a function depending on the arguments

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+ * @template HorizontalDomain
+ * @template VerticalDomain
+ * @param {d3.Selection<GElement, Datum, PElement, PDatum>} d3Instance
+ * @param {Array<GeoChart.SingleStackedBarGroupConfig<HorizontalDomain, VerticalDomain>>} options
+ * @param {GeoChart.GlobalOptions} globalOptions
+ * @param {d3.Tooltip<SVGElement, Datum, PElement, PDatum>} [d3TipInstance]
+ */
+export function render (d3Instance, options, globalOptions, d3TipInstance) {
+  const groups = d3Instance
+    .selectAll('g.geo-chart-stacked-bars-group')
```
fofi: We could move the class to it's own variable, so we do not repeat it 6 times

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+    })
+  }
+
+  function getBarHeight (d, i) {
+    switch (singleGroupOptions.mainDimension) {
+      case dimensionUtils.DIMENSIONS_2D.horizontal:
+        return getWidthTranslation(d, i)
+      case dimensionUtils.DIMENSIONS_2D.vertical:
+        return Math.abs(
+          _.first(axisForMainDimension.scale.axisScale.domain()) - _.last(axisForMainDimension.scale.axisScale.domain())
```
fofi: We could just get the domain once and use it in both functions

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+      default:
+        console.error(`GeoChartStackedBars [component] :: Invalid axis main dimension for getBarInitialWidth: ${singleGroupOptions.mainDimension}`)
+    }
+  }
+
+  function getBarWidth (d, i) {
+    switch (singleGroupOptions.mainDimension) {
+      case dimensionUtils.DIMENSIONS_2D.horizontal:
+        return Math.abs(
+          _.first(axisForMainDimension.scale.axisScale.domain()) - _.last(axisForMainDimension.scale.axisScale.domain())
```
fofi: We could just get the domain once and use it in both functions

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+  isMainDimensionHorizontal
+}, d3TipInstance) {
+  const segmentBaseClass = 'geo-chart-stacked-bars__segment'
+
+  const segmentsData = stackedBarData[axisForMainDimension.keyForValues]
+
+  let temporaryBarWrapperHeight = 0
+  segmentsData.forEach((segment, idx) => {
+    segment.startValue = idx === 0
+      ? _.min(axisForMainDimension.scale.axisScale.domain())
```
fofi: Here we are getting the domain 2 times per loop, I think it wont change for each loop, so we can just take it once before the forEach and use it inside

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+  axisForNormalDimension,
+  isMainDimensionHorizontal
+}, d3TipInstance) {
+  const segmentBaseClass = 'geo-chart-stacked-bars__segment'
+
+  const segmentsData = stackedBarData[axisForMainDimension.keyForValues]
+
+  let temporaryBarWrapperHeight = 0
+  segmentsData.forEach((segment, idx) => {
+    segment.startValue = idx === 0
```
fofi: if we initialize temporaryBarWrapperHeight to _.min(domain()), we really dont have to care about idx === 0

src/elements/GeoChart/GeoChartStackedBars/GeoChartStackedBars.js
```diff
@@ -0,0 +1,363 @@
...
+      ? _.min(axisForMainDimension.scale.axisScale.domain())
+      : temporaryBarWrapperHeight
+    segment.endValue = idx === segmentsData.length - 1
+      ? _.max(axisForMainDimension.scale.axisScale.domain())
+      : temporaryBarWrapperHeight + segment[axisForMainDimension.keyForValues]
```
fofi: in which cases these 2 lines were returning different for the last segment?

src/elements/GeoChart/GeoChartUtils/axisUtils.js
soleneG02: Because in this case we don't have a prop `naturalOffset` in the config, and we don't really want a customized offset, the goal is just to place the bar under it's legend. Maybe I should rename the variable

PR #22, opened 2020-01-24
Feature/core 7901 scatter plot

soleneG02:


src/elements/GeoChart/GeoChartConfigs/GeoChartConfig.js
```diff
@@ -733,6 +733,100 @@ export const anchoredShapesConfigSchema = {
...
+    // Note that there might be some of the default classes might be added
+    // regardless to your customization as they are required internally.
+    verticalAxisConfig: axisConfigJsonSchema,
+    // Function taking as first parameter an array of CSS classes that would be
+    // set by default. Should return the array of CSS classes to be finally set.
+    // Use this function to customize which CSS classes are set to each line.
+    // Note that there might be some of the default classes might be added
+    // regardless to your customization as they are required internally.
+    horizontalThreshold: {
+      type: ['number', 'string']
```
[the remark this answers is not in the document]
soleneG02: In the case of a `categoricalAxis`, the value passed as threshold is the string of the legend, so I had to add this type 'string'

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -0,0 +1,295 @@
...
+    const isAlreadyClickedDot = clickedDot.indexSingleGroup === index && clickedDot.indexDot === i
+    if (isAlreadyClickedDot) return
+
+    d3.select(this)
+      .style('fill', _.defaultTo(singleGroupOptions.fillColor, DEFAULT_FILL_COLOR))
+      .style('stroke', '#EBEBEB')
+      .style('stroke-width', '2')
+  }
+
+  function handleClick (d, i) {
```
dpramar: I don't like much how we do this, having the object `clickedDot`.

Do you think that can be in each dot? like when a dot is clicked add a property `d.isDotSelected: !d.isDotSelected` then to the rest of dots reset `d.isDotSelected: false`

soleneG02: I can do that but in this case, we have to check for every dots of the chart when clicking on a new one and I though that would be a big loop, so I tried to avoid it

soleneG02: And I think when I tried, another problem was accessing the previous clicked dot to change its style within the `handleClick` function..

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -0,0 +1,313 @@
...
+      d.isClicked = true
+
+      return singleGroupOptions.onDotClick(d, i)
+    } else if (isAlreadyClickedDot) {
+      const color = _.defaultTo(singleGroupOptions.getFillColor(d, i), DEFAULT_FILL_COLOR)
+      unclickedStyle(d3.select(this), color)
+      d.isClicked = false
+
+      return singleGroupOptions.onDotClick(null, null)
+    } else if (!!previouslyClickedDot && !isAlreadyClickedDot) {
```
fofi: We dont really need this if else, as here we already know thanks to the previous conditions that !isAlreadyClickedDot and that is previouslyClickedDot

The code wont neither enter here as we have the return statements above.
We could also split the other condition due this

This will make the code a little bit more readable as it has less conditionals

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
fofi: I think this function is quite difficult to read. Maybe we could split it in smaller functions. We could sit together tomorrow

src/elements/GeoChart/GeoChartQuadrant/GeoChartQuadrant.js
```diff
@@ -0,0 +1,351 @@
...
+ * @template Domain
+ * @template RelativeScaleDomain
+ * @param {d3.Selection<GElement, Datum, PElement, PDatum>} group
+ * @param {d3.Tooltip<SVGElement, object, PElement, PDatum>} [d3TipInstance]
+ * @param {Array<GeoChart.SingleQuadrantGroupConfig<Domain, RelativeScaleDomain>>} singleQuadrantOptions
+ * @param {GeoChart.GlobalOptions} globalAxesConfig
+*/
+function renderSingleQuadrant (group, d3TipInstance, singleQuadrantOptions, globalAxesConfig) {
+  const showVerticalLine = singleQuadrantOptions.horizontalThreshold !== undefined
+  const showHorizontalLine = singleQuadrantOptions.verticalThreshold !== undefined
```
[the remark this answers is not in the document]
soleneG02: Problem is that with `!!singleQuadrantOptions.verticalThreshold`, `0` is false even though it should pass
But I can do `!!singleQuadrantOptions.verticalThreshold || singleQuadrantOptions.verticalThreshold === 0` ? I don't know if it's better

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -0,0 +1,313 @@
...
+    .exit()
+    .transition()
+    .duration(globalOptions.chart.animationsDurationInMilliseconds)
+    .style('opacity', 0)
+    .remove()
+
+  const updatedGroups = groups
+  const allGroups = newGroups.merge(updatedGroups)
+
+  const clickedDot = {
```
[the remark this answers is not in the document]
soleneG02: I forgot to remove the definition sorry.. It's been replaced with the isClicked prop on each dot now

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
soleneG02: I don't understand what you mean, let's see together today for the whole function if you have time

src/elements/GeoChart/GeoChartQuadrant/GeoChartQuadrant.js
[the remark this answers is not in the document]
fofi: We can try with `_.isFinite()`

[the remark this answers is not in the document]
soleneG02: Doesn't work either because the `threshold` can be a string (when we have a categorical axis)

src/elements/GeoChart/GeoChart.vue
```diff
@@ -306,6 +320,7 @@ export default {
       this.adjustSize()
       this.redrawGuidelines()
       this.updateData()
+      this.redrawQuadrants()
```
jmenchero: Does it makes sense to have this inside redrawAxes?

src/elements/GeoChart/GeoChartQuadrant/GeoChartQuadrant.d.ts
```diff
@@ -0,0 +1,18 @@
...
+      quadrantBottomLeftName?: string
+      quadrantBottomRightName?: number
+      fontSize?: number
+      tooltip?: {
+        getContent?: (item: object, index: number) => string | null | undefined
+        getOffset?: (event: MouseEvent) => { x: number, y: number } | null | undefined
+      }
+      cssClasses?: (defaultClasses: string[], item: object, index: number) => string[]
+    }
+  }
```
jmenchero: Would add an extra empty line.

src/elements/GeoChart/GeoChart.vue
soleneG02: I don't think it changes much since the whole `redraw` function is called anytime one of them is changing (and also quadrant can be changed when axis aren't like a change of threshold for example)

PR #25, opened 2020-02-10
Feature/core 7953 geo calendar day name alignment fix

Navas23:


src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
```diff
@@ -87,9 +87,12 @@
   .geo-calendar-grid__weekdays-row-container {
     @include font($label-optional);
 
+    display: grid;
```
jmenchero: I would like to ask opinion to @soleneG02 and @fofi about grid, is it used across the DS? If it is not the standard placement method I would go to properly tweak the flex method.

src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
```diff
@@ -87,9 +87,12 @@
...
+    display: -ms-grid;
```
jmenchero: I am not a big fan of using values that require vendor prefixes to work, would try to avoid them if possible.

src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
```diff
@@ -87,9 +87,12 @@
...
+    grid-template-columns: repeat(7,1fr);
+    -ms-grid-template-columns: repeat(7,1fr);
```
jmenchero: Same with the grid-template-columns attribute. Also, isn't splitting it into 7 fractions too hardcoded?

Navas23: It is, but that's the way the calendar content is using too...Maybe we can make a new issue to upgrade calendar to be able to work with variable column size?

src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
Navas23: Me neither, I used it to unify styles with the calendar content

src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
fofi: I have no strong opinion on this, use the solution that the team think is better

src/styles/elements/GeoCalendar/_geo-calendar-grid.scss
jmenchero: Sounds fair enough!

PR #26, opened 2020-02-10
fix scatter plot and quadrant & add prop for axes

soleneG02:
- use class `geo-chart-axis--with-quadrant` only in demo (and not always) to be able to customize it in the app
- fix label position for `GeoChartQuadrant`
- add class to handle clicking on a scatter plot dot from an external component
- add property to force `ticks` number in axis (in the previous way, d3 rounded the number of ticks to a multiple of 2, 5, or 10 when creating ticks and so didn't accept all values)

### DIFF
--- a/src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
+++ b/src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
@@ -415,9 +415,10 @@ export function createCustomizedTickArray (firstOfDomain, lastOfDomain, tickCoun
   const domainRange = Math.abs(firstOfDomain - lastOfDomain)
   const ticksStep = domainRange / (tickCount - 1)
   const tickToDisplay = _.map(_.times(tickCount), (id) => {
-    return firstOfDomain < lastOfDomain
+    return _.round(firstOfDomain < lastOfDomain
       ? firstOfDomain + (ticksStep * id)
       : firstOfDomain - (ticksStep * id)
+    , 3)
   })
   return tickToDisplay
 }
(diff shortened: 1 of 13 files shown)


src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
```diff
@@ -402,3 +410,15 @@ function positionLabel (label, singleAxisOptions, globalAxesConfig) {
       break
   }
 }
+
+function createCustomizedTickArray (firstOfDomain, lastOfDomain, tickCount) {
```
fofi: Do we have tests for the graphs? we should probably also test this small functions

test/unit/specs/elements/GeoChart/GeoChartAxis.spec.js
```diff
@@ -369,6 +369,59 @@ describe('GeoChartAxis', function () {
...
+        expect(wrapper.find('.geo-chart-axis .tick').exists()).toBe(true)
+        expect(wrapper.findAll('.geo-chart-axis .tick')).toHaveLength(tickCount + 1)
+
+        expect(wrapper2.find('.geo-chart').exists()).toBe(true)
+        expect(wrapper2.find('.geo-chart-axis').exists()).toBe(true)
+        expect(wrapper2.find('.geo-chart-axis .tick').exists()).toBe(true)
+        expect(wrapper2.findAll('.geo-chart-axis .tick')).toHaveLength(tickCount)
+
+        const ticksArray = [100, 75, 50, 25, 0]
+        expect(GeoChartAxis.createCustomizedTickArray(startDomain, endDomain, tickCount)).toMatchObject(ticksArray)
```
fofi: We should add a test just for this function.
In that test we should check multiple possible inputs, for example bigger start domain that end domain, or equal, or negative, etc..

PR #29, opened 2020-02-24
Feature/core 8000 geo scatter plot dot coordinates

Navas23:
GeoChartScatterPlot returns coordinates on click event


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 29.5.0
```
fofi: Should this change be a minor or a patch?

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -255,8 +255,10 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
       unclickedStyle(dotToUnclick, previouslyClickedDot)
     }
 
-    clickedStyle(d3.select(this), d)
-    return singleGroupOptions.onDotClick(d, d.index)
+    const clickedDot = d3.select(this)
+    const dotCoordinates = [parseFloat(clickedDot.attr('cx')), parseFloat(clickedDot.attr('cy'))]
```
fofi: Is this the absolute screen value? I think we should add a test regarding this if possible

CHANGELOG.md
Navas23: To me (maybe I'm wrong) a minor is used just for bugfixes, that's why I decided to go for a patch.

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
Navas23: Yes, adding a test make sense even if it should never fail as the dot is already in the DOM

fofi: I mean, I think we should verify that the returned value is the value we expect.
Imagine that we are getting the value inside the graph container but not the absolute one to the entire screen

Navas23: Ok, I didn't get you the first time. The coordinates are relative to the GeoChart component, is the Geoblink app who decides what to do with the coordinates... Would you add a test anyway? (I think is not needed bc it should work 100%)

PR #32, opened 2020-03-03
Feature/core 8082 not interactive scatter plot

soleneG02:


src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -228,6 +266,7 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
   }
 
   function handleMouseOut (d, i) {
+    if (!singleGroupOptions.tooltip) return
```
fofi: what does the tooltip has to do with the mouse over/out?

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -141,19 +142,30 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
...
     .on('mouseout', handleMouseOut)
     .on('click', handleClick)
+    .classed('is-clicked', (d) => classed(d))
     .transition()
     .duration(globalOptions.chart.animationsDurationInMilliseconds)
-    .style('stroke', 'white')
-    .style('stroke-width', '2')
+    .style('stroke', (d) => stroke(d))
+    .style('stroke-width', (d) => strokeWidth(d))
+    .style('stroke-opacity', (d) => strokeOpacity(d))
```
fofi: Should we have a function to set all the style props? I think in d3 you can use `.styles`

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -165,6 +177,31 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
 
   setupTooltipEventListeners(allDots, d3TipInstance, singleGroupOptions.tooltip)
 
+  function classed (d) {
+    if (d.index === singleGroupOptions.initiallyClickedIndex) {
+      d.isClicked = true
```
fofi: I think we should not modify anything in something that should just set a class

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -165,6 +177,31 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
...
+    }
+    return d.index === singleGroupOptions.initiallyClickedIndex
+  }
+
+  function stroke (d) {
+    return d.index === singleGroupOptions.initiallyClickedIndex
+      ? '#9B9B9B'
```
fofi: should this be part of the DS? if so, we should use a constant like we are doing with DEFAULT_FILL_COLOR

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
soleneG02: When there is no tooltip or no action on click (meaning no action at all) on a dot, I don't think we should change the form of the dot, but actually, the condition should be `!singleGroupOptions.tooltip && !singleGroupOptions.onDotClick`

fofi: but, maybe we will want just visual feedback but no tooltip at all, I think we should allow it on that case

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
soleneG02: I didn't found anything related to .styles, but I'll create a function to have .attr('style', (d) => applyStyle(d))

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
soleneG02: I'll add a prop then!

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -177,29 +178,48 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
 
   setupTooltipEventListeners(allDots, d3TipInstance, singleGroupOptions.tooltip)
 
+  function applyStyle (d) {
+    const color = `fill: ${_.defaultTo(singleGroupOptions.getFillColor(d, d.index), DEFAULT_FILL_COLOR)};`
+    return `${color} ${cursor(d)} ${opacity(d)} ${stroke(d)} ${strokeWidth(d)} ${strokeOpacity(d)}`
+  }
+
```
fofi: There is something called [d3-selection-multi](https://github.com/d3/d3-selection-multi#d3-selection-multi), maybe we can use it here

soleneG02: This dependency (it's not in the default package) requires to have a higher version of node than the one we have now (8.. to 10.. at least), and I saw with Juan that it's not something straightforward (there was a task for upgrading node but you put it to won't address)
Do you think it's really important to have this dependency ?

fofi: No, better not to have it.

Maybe we could use [this](https://github.com/d3/d3-3.x-api-reference/blob/master/Selections.md#style) directly from d3

`selection.style({stroke: "black", "stroke-width": "2px"});
`

soleneG02: It doesn't work either, the error is saying that `style` doesn't take an object as parameter. And in forums, they're always saying we should use `styles` with d3-selection-multi for it to work (https://stackoverflow.com/questions/41474452/style-method-does-not-take-object)

fofi: Oh, it was removed on v4.. Sorry for the noise

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -165,6 +178,50 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
 
   setupTooltipEventListeners(allDots, d3TipInstance, singleGroupOptions.tooltip)
 
+  function applyStyle (d) {
+    const color = `fill: ${_.defaultTo(singleGroupOptions.getFillColor(d, d.index), DEFAULT_FILL_COLOR)};`
```
fofi: why not having a function also for the color?

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -165,6 +178,50 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
...
+    return `${color} ${cursor(d)} ${opacity(d)} ${stroke(d)} ${strokeWidth(d)} ${strokeOpacity(d)}`
```
fofi: I think it would be great to move all these functions inside the applyStyle just to keep clear the scope of them,

src/elements/GeoChart/GeoChartScatterPlot/GeoChartScatterPlot.js
```diff
@@ -113,7 +128,7 @@ function renderSingleGroup (group, d3TipInstance, singleGroupOptions, globalOpti
 
   _.each(singleGroupOptions.data, (dot, i) => {
     dot.index = i
-    dot.isClicked = false
+    dot.isClicked = singleGroupOptions.initiallyClickedIndex === i
```
fofi: are we sure that we are just executing this one time when initializiating?

soleneG02: No you're right it will be added everytime the graph is updated.. Would it be "clean" to create a function `initScatterPlotGroups` in `GeoChartConfigAdapter.scatterPlot.mixin.js`, or to pass an argument to `updateScatterPlotGroups` to specify if it's the initialization or an update ?

PR #36, opened 2020-03-10
add animationDuration prop to geoScatterPlot

soleneG02:
In the app for SA2.0, we need the animation for the dots to be independent from the axis, guidelines and quadrant ones (because we don't want any animation on update for the scatterPlot dots)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.5.3",
+  "version": "29.5.4-beta.0",
   "description": "Geoblink Design System for Vue.js",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.5.3",
+  "version": "29.5.4-beta.0",
```
jmenchero: Was the master version set to a beta?? Shouldn't that be avoided?

soleneG02: Just a beta until it's approved to be merged, this way I can work in the app in parallel, but I'll change it right before merging!

PR #37, opened 2020-03-17
Feature/core 8070 create kpi component

Navas23:


src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <div
+    class="geo-KPI"
```
fofi: i do not really like having uppercase css classes, what do you think team?

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
...
+    :class="[`geo-KPI${valueCSSSuffix}`,{
+      'geo-KPI--is-primary' : data.isPrimary,
+      'geo-KPI--is-secondary' : !data.isPrimary,
+    }]"
```
fofi: why not adding all the classes the same way? not sure why we need the brackets `{}` here

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
...
+      validator (data) {
+        return _.isBoolean(data.isPrimary) && data.value
+      }
+    }
+  },
+  data () {
+    return {
+
+    }
+  },
```
fofi: I think there is no need in setting this as empty

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
...
+]
+
+export default {
+  name: 'GeoKPI',
+  status: 'ready',
+  release: '29.6.0',
+  props: {
+    /**
+     * Object containing all the KPI data
+     *
```
fofi: I think we could add here the properties supported. something similar to jsdocs?

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
...
+    }
+  },
+  data () {
+    return {
+
+    }
+  },
+  computed: {
+    valueCSSSuffix () {
+      return this.data.colorHighlight && _.includes(COLOR_MODIFIERS, this.data.colorHighlight) ? `--${this.data.colorHighlight}` : ''
```
fofi: I think it would be enough just with the includes, as we do not have anything falsy inside that object

fofi: actually, we could validate in the input that `colorHighlight` has a valid modifier

src/styles/elements/GeoKPI/_geo-kpi.scss
```diff
@@ -0,0 +1,57 @@
+
+.geo-KPI{
+  border-radius: $border-radius-default;
+}
+
+.geo-KPI__column{
+  float: left;
+}
+
+.geo-KPI--is-primary .geo-KPI__value, .geo-KPI--is-primary .geo-KPI__unit{
```
fofi: if we want to do nesting, I think is more readable using several lines.
eg
```
.geo-KPI--is-primary {
  .geo-KPI__value, .geo-KPI__unit {
    @include font($display-medium);
    color: #545454;
    display: inline-block;
  }
  .geo-KPI__description{
    @include font($body-small);
    color: $color-grey-darker;
    margin-top: 3px;
  }
}
```

src/styles/elements/GeoKPI/_geo-kpi.scss
```diff
@@ -0,0 +1,57 @@
...
+  @include font($display-medium);
+  color: #545454;
```
fofi: lets try to use always colors from the color file. that way is easier to use always the same palette and to update it

src/styles/elements/GeoKPI/_geo-kpi.scss
```diff
@@ -0,0 +1,57 @@
...
+  display: inline-block;
+}
+
+.geo-KPI--is-primary .geo-KPI__description{
+  @include font($body-small);
+  color: $color-grey-darker;
+  margin-top: 3px;
+}
+
+.geo-KPI--is-primary.geo-KPI--red .geo-KPI__value, .geo-KPI--is-primary.geo-KPI--red .geo-KPI__unit{
```
fofi: I think the modifier should be in `.geo-KPI__value` and in `.geo-KPI__unit`. not in the father.

src/styles/elements/GeoKPI/_geo-kpi.scss
fofi: probably the same here, why is the modifier in the parent?

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,82 @@
...
+    }
+  },
+  methods: {
+
+  }
```
fofi: same here, we dont need an empty definition here

test/unit/specs/elements/GeoKPI/GeoKPI.spec.js
```diff
@@ -0,0 +1,32 @@
+import { mount } from '@vue/test-utils'
+import GeoKPI from '@/elements/GeoKPI/GeoKPI.vue'
+
+describe('GeoKPI', () => {
+  it('Should display GeoKPI primary value', () => {
```
fofi: I think we should test all the possible scenarios, sending colors, wrong colors, units, etc.

src/styles/elements/GeoKPI/_geo-kpi.scss
Navas23: I don't like to hardcode the color but it's the one I got on Invision, gonna talk with Gui about it

src/styles/elements/GeoKPI/_geo-kpi.scss
Navas23: Think that modifier do different things depending of isPrimary value... you have 2 approaches:

- Decide in the template where to add the css modifiers depending of isPrimary value
- Add the modifier to the parent and use it wherever you need it in css (unit and value in case of isPrimary == true and parent container if isPrimary == false)

I went for the last approach, do you prefer the first one?

src/elements/GeoKPI/GeoKPI.vue
Navas23: AFAIK we need to declare data if we want it to be available for use in the slots

src/elements/GeoKPI/GeoKPI.vue
Navas23: Yes, I'll modify all to lowercase

src/elements/GeoKPI/GeoKPI.vue
fofi: but, do slots use this data, or the parents data?

src/styles/elements/GeoKPI/_geo-kpi.scss
fofi: I think we should follow the same approach in all the components

src/elements/GeoKPI/GeoKPI.vue
Navas23: You are totally right, parents data

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -53,30 +49,35 @@ export default {
...
      *
-     * **Note:** value and isPrimary properties are required, check examples to get more details.
      */
     data: {
       type: Object,
       required: true,
       validator (data) {
-        return _.isBoolean(data.isPrimary) && data.value
+        if (!_.isBoolean(data.isPrimary)) return false
+        if (!data.value) return false
```
fofi: Are we going to allow falsy values? falsy = 0, null

Navas23: I don't think so but I will modify it in order to be more "robust"

src/styles/elements/GeoKPI/_geo-kpi.scss
```diff
@@ -0,0 +1,67 @@
...
+
+.geo-kpi--is-primary{
+  .geo-kpi__value, .geo-kpi__unit{
+    @include font($display-medium);
+    color: $color-grey-darker;
+    display: inline-block;
+  }
+}
+
+.geo-kpi--is-primary .geo-kpi__description{
```
fofi: lets try to unify as much as possible all the classes under the same modifier together

src/styles/elements/GeoKPI/_geo-kpi.scss
```diff
@@ -0,0 +1,67 @@
...
+  @include font($body-small);
+  color: $color-grey-darker;
+  margin-top: 3px;
+}
+
+.geo-kpi--is-primary.geo-kpi--red{
+  .geo-kpi__value, .geo-kpi__unit{
+    color: $color-red;
+  }
+}
```
fofi: I am not sure about this, but I think to unify this part we could do something like this
```
.geo-kpi--is-primary {
  &.geo-kpi--red {
    .geo-kpi__value, .geo-kpi__unit {
      color: $color-red;
  }
}
```

src/styles/elements/GeoKPI/_geo-kpi.scss
fofi: just a minor thing, I think we should also keep the consistency when writing classes. eg: keeping an space between the class and the bracket

src/styles/elements/GeoKPI/_geo-kpi.scss
Navas23: The modifier is in the parent because it affects all the KPI component, and depending of the type of the KPI data is going to modify different elements of the component

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,83 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+/**
+ * `GeoKPI` is a component for displaying data, specially designed to display Key Performance Indicators
+ */
+import _ from 'lodash'
+const COLOR_MODIFIERS = [
```
fofi: Confirm with juan the usage of the modifiers. I think we should use them like we are doing in the geoAlert or in the geoButton.

They are more "types" than modifiers. We could export them and the usage would be just sending the type you want

src/elements/GeoKPI/GeoBadKPI.vue
```diff
@@ -0,0 +1,26 @@
...
+ * See [GeoKPI](./GeoKPI) for a complete list of supported properties and
+ * features.
+*/
+export default {
+  name: 'GeoBadKPI',
+  status: 'ready',
+  release: '29.6.0',
+  mixins: [mixin],
+  computed: {
+    isPrimary () {
```
fofi: should we add this property to the mixing? I think is in all the components!

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
...
+     * [data.unit] {string} - KPI data unit
+     *
+     * [data.description] {string} - KPI data description
+     *
+     */
+    data: {
+      type: Object,
+      required: true,
+      validator (data) {
+        if (!_.isBoolean(data.isPrimary)) return false
```
fofi: didn't we changed this and move it to its own property?

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
...
+     * data.isPrimary {Boolean} - Determines if KPI data is primary or not
+     *
+     * data.value {string|Number} - KPI data value
+     *
+     * [data.unit] {string} - KPI data unit
+     *
+     * [data.description] {string} - KPI data description
+     *
+     */
+    data: {
```
fofi: one question regarding this property, why not adding several ones? I meam, we dont really need an object, we can add 4 properties

Navas23: I started developing it as 4 props and I switched to 1 object because it made more sense to me (conceptually). What do you think?

src/elements/GeoKPI/GeoBadKPI.vue
Navas23: Yes, probably makes more sense to have it in the mixin

src/elements/GeoKPI/GeoKPI.mixin.js
Navas23: I dont get what you mean here

src/elements/GeoKPI/GeoKPI.mixin.js
fofi: I think that what we are actually doing in general for similar components is to set several props

src/elements/GeoKPI/GeoKPI.mixin.js
fofi: I though that we've moved `isPrimary` to its own property, not inside data

Navas23: We didn't, but as we are talking I think we will change from object to several props approach

src/elements/GeoKPI/GeoStandardKPI.vue
```diff
@@ -0,0 +1,26 @@
+<template src="./GeoKPI.template.html" />
```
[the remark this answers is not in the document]
Navas23: Makes sense to me

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,81 @@
...
+import { enumPropertyFactory } from '../../utils/enumPropertyFactory'
+
+/**
+ * `GeoKPI` is a component for displaying data, specially designed to display Key Performance Indicators
+ */
+
+export default {
+  name: 'GeoKPI',
+  status: 'ready',
+  release: '29.6.0',
```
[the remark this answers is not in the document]
Navas23: I think it is just informational

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash'
+
+const TYPES = {
+  good: 'good',
```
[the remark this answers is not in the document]
Navas23: This is outdated

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
...
+  medium: 'medium',
+  bad: 'bad',
+  standard: 'standard'
```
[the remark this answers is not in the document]
Navas23: This is outdated too hahah

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export { TYPES }
+/**
+ * @mixin
```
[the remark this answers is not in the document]
Navas23: As we want to be able to use directly "GeoPositiveKPI" and stuff like that mixin is the only way to not duplicate code

src/elements/GeoKPI/GeoKPI.mixin.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash'
+
+const TYPES = {
```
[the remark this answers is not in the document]
Navas23: Which benefits do you see to your approach?

src/elements/GeoKPI/GeoKPI.template.html
```diff
@@ -0,0 +1,13 @@
+<geo-KPI
+  v-bind="$props"
+  :type="type"
+>
+  <slot
+    slot="warningTooltip"
+    name="warningTooltip"
+  />
+  <slot
```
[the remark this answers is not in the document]
Navas23: The template seem the best way to unify code between all the KPIs. And slots are the only way if we want to allow custom icons

src/elements/GeoKPI/GeoKPI.vue
```diff
@@ -0,0 +1,81 @@
+<template>
```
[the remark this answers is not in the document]
Navas23: Idk...

src/elements/GeoKPI/GeoKPI.vue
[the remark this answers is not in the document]
fofi: I think it for the docs, there you can see the release of each component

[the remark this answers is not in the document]
Navas23: Yes it is, i think that's what he means with "informational" purposes

src/elements/GeoKPI/GeoKPI.mixin.js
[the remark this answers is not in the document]
jmenchero: Do we want to? I think positiveness or negativeness of a component is related to its value, not as something to be forced from outside to be positive or negative despite the value. And nevertheless, in case we want to force the colour, style modifiers should be passed as props, I think having a component per colour is an overkill in case we, or anyone (we have to remember this is opensource) wants more colours, it will have to grow up to n colours, and also it doesn't feel like the expected behaviour of a component, the standarized way to handle colours in components is with external styling or props.

src/elements/GeoKPI/GeoKPI.mixin.js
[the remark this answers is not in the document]
jmenchero: I think a prop "colour" that recives either a colour, or a function that returns a colour based on the value, makes it more customizable, delegates the logic of labeling the data to the parent. It allows flexibility.

src/elements/GeoKPI/GeoKPI.template.html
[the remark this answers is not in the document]
jmenchero: I think the mixin approach makes the component harder to maintain, as result of all this kind of workarounds, while it doesn't add any value to my point of view, but actually makes the component less flexible.

src/elements/GeoKPI/GeoKPI.vue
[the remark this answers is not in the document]
jmenchero: The KPI naming is suggested from product as the higher level component, the one that the users are going to use in the dashboards (which might include benchmarking or even multidimensionality KPIs). To put it in an atomic design categorization, this would be an atom that will be part of the KPI molecule, so we shouldn't mix namings in order to avoid confusion.

src/elements/GeoKPI/GeoKPI.vue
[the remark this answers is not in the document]
jmenchero: :ok_hand:

src/elements/GeoKPI/GeoKPI.mixin.js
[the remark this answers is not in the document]
fofi: this is not really something related to colors, that they could also be overwritten from the outside.
Here we are exposing an api to create a positive or negative behavior of the component. We are actually doing the same for buttons, danger, link, etc.

I think we should merge this now and open a new discussion if we want to start to do things different

src/elements/GeoKPI/GeoKPI.mixin.js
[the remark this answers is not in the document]
fofi: This is not something color related, but behavior.
I dont think we should send as parameter a function that returns a color just to get a simple color in such a simple component.

again, lets discuss this in other place if we want to change how we are doing things

src/elements/GeoKPI/GeoKPI.vue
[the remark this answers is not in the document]
fofi: agree with this but I cannot think on a better name now.
I think we should close this PR as soon as possible.
Is not ideal btu we could rename it later

src/elements/GeoKPI/GeoKPI.template.html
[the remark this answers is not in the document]
Navas23: I think this is more a decision about splitting the component in 4 like we are doing now or not... Maybe we can let it be like this and modify it later if we don't like the approach when we need to use it

src/elements/GeoKPI/GeoKPI.vue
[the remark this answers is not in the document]
jmenchero: There is nothing more permanent than a temporary solution hahaha. And I would say, as this is the first stone of the new experience, we should put particular care in taking this kind of important decissions, as they are gone to set the tone of the whole new development. I don't have an strong opinion on this but I don't see any motivation to keep the KPI name while knowing it is wrong or the component per colour while knowing it is not the common usage case, neither flexible. Nevertheless I am open to hear why you both think it is the best approach to keep it like this.
This is also the same motivition for the other discussion.

PR #65, opened 2020-07-10
Feature/core 8228 fix geocalendar selection

Navas23:


src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -392,9 +394,7 @@ export default {
         ? unverifiedRangeSettings.whenSettingFromDate
         : unverifiedRangeSettings.whenSettingToDate
 
-      const isRangeValid = unverifiedRange.end
-        ? isBefore(unverifiedRange.start, unverifiedRange.end)
-        : true
+      const isRangeValid = !unverifiedRange.end || (!unverifiedRange.start || isBefore(unverifiedRange.start, unverifiedRange.end))
```
asanzdj, line 397: This code block is repeat, it could be nice to extracted it into a method to reuse it in two sides. What do you think about it?

Navas23: I agree! On my way

src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -569,6 +567,10 @@ export default {
       }
     },
 
+    validateRange (start, end) {
+      return !end || (!start || isBefore(start, end))
```
fofi, line 571: I feel like its a little bit confusing this condition. maybe turning it around?
`
isInvalid = start && end && !isBefore(start, end)
isValid = !(start && end && !isBefore(start, end))
`
or more similar to what we had
`
isValid = start && end ? isBefore(start, end) : true
`

Navas23: I like the 2nd one!

PR #69, opened 2020-09-14
Feature/core 7973 clean up test console errors

asanzdj:
**Jira task:**

https://geoblink.atlassian.net/browse/CORE-7973

**App client PR**

https://bitbucket.org/geoblink/geoblink-app/pull-requests/6438

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.8.3-beta.4",
+  "version": "29.8.4",
   "description": "Geoblink Design System for Vue.js",
   "author": "Geoblink <contact@geoblink.com>",
   "main": "dist/system.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.8.2",
+  "version": "29.8.3-beta.1",
   "description": "Geoblink Design System for Vue.js",
   "author": "Geoblink <contact@geoblink.com>",
   "main": "dist/system.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.8.3-beta.3",
+  "version": "29.8.3",
   "description": "Geoblink Design System for Vue.js",
   "author": "Geoblink <contact@geoblink.com>",
   "main": "dist/system.js",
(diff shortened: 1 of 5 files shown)


fofi (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.8.2",
+  "version": "29.8.3-beta.3",
```
Navas23, line 3: Try to open the PR using stable version

Navas23 (commented):
I'm missing the changelog changes. Appart from that, have you tested possible side effects of this?

src/elements/GeoSegmentedControl/GeoSegmentedControlItem.vue
```diff
@@ -18,8 +18,10 @@
 </template>
 
 <script>
+import GeoTrimmedContent from '@/elements/GeoTrimmedContent/GeoTrimmedContent.vue'
```
Navas23, line 21: Where are you using GeoTrimmedContent?

asanzdj: In the component, you can find it at line 13

Navas23: I think you don't need to import DS components here

asanzdj: Ah, I made it because testing was complaining and I found that it wasn't importing it, but, I'm going to try it again

Navas23: Maybe you will need to import it in the tests, I'm not 100% sure but as I see no changes in the template I guess this was working without importing it

asanzdj:
> I'm missing the changelog changes. Appart from that, have you tested possible side effects of this?

Before and after changing it I ran test in DS and all was right.  I suppose this problem happens applying it into a real app, because I found the problem occurs when component is updating (on $nextTick) and maybe at that moment the DOM is not completely loaded.

src/elements/GeoSegmentedControl/GeoSegmentedControlItem.vue
asanzdj: You're right. It is not necessary.

fofi (approved):

Navas23 (approved):

PR #73, opened 2020-09-29
(feat) Starting new geo-tree component

asanzdj:
Jira task:

https://geoblink.atlassian.net/browse/CORE-8455


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,107 @@
...
+    <span
+      v-if="hasChildren"
+      class="geo-tree-item__total-items"
+    >
+      ({{ totalItems }})
+    </span>
+    <template slot="trailingAccessoryItem">
+      <div v-if="category.description">
+        <font-awesome-icon
+          :icon="['fal', 'info-circle']"
```
asanzdj, line 18: Where I have to import this component?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,215 @@
...
+            <slot name="actionButton" />
+          </template>
+        </geo-tree-item>
+      </ul>
+    </geo-scrollable-container>
+  </div>
+</template>
+
+<script>
+import _, { sortBy, assign, omit, includes, isEmpty } from 'lodash'
```
Navas23, line 40: Not sure if it's better or not but... why not simply import _ and then use _.whatever you want to use?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,215 @@
...
+     *     children: [
+     *      id,
+     *      label,
+     *      children
+     *     ]
+     * ]
+     */
+    categories: {
+      type: Array,
+      default: () => []
```
Navas23, line 76: Can't we use directly [ ] ?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,215 @@
...
+      filteredCategories: this.categories
+    }
+  },
+  computed: {
+    formattedCategories () {
+      return sortBy(this.filteredCategories, [this.keyForLabel])
+    }
+  },
+  watch: {
+    initialState: {
```
Navas23, line 132: Why do we need a watcher for this?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,215 @@
...
+      },
+      deep: true
+    }
+  },
+  mounted () {
+    this.setInitialState(this.initialState)
+  },
+  methods: {
+    handleSearching () {
+      this.$emit('search', this.searchQuery)
```
Navas23, line 144: I don't clearly understand how are you handling the search box... As soon as you finish this we can set a small meeting so i get it

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Changed to use prop

src/elements/GeoTree/GeoTree.vue
asanzdj: In new version I'm using _ according to the lodash agreement

src/elements/GeoTree/GeoTree.vue
asanzdj: No, eslint complains, but I'm not sure why with arrays is needed too (I am not able to find an explanation). https://eslint.vuejs.org/rules/require-valid-default-prop.html

src/elements/GeoTree/GeoTree.vue
asanzdj: Nope, you're right

src/elements/GeoTree/GeoTree.vue
asanzdj: This is hard, I'm working on it

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,226 @@
...
+    descriptionIcon: {
+      type: Array,
+      default: () => []
+    }
+  },
+  data () {
+    return {
+      searchQuery: '',
+      isLoading: false,
+      checkedItems: {},
```
fofi, line 139: Do you think that we should have the state here or in the father (the one creating this component)?

src/styles/elements/GeoTree/_geo-tree-item.scss
```diff
@@ -0,0 +1,26 @@
...
+  width: 5px;
+}
+
+.geo-tree-item__total-items {
+  @include font($body-small);
+  color: $color-grey-light;
+  font-style: italic;
+}
+
+.geo-list-item__icon-and-label__icon-container {
```
asanzdj, line 20: I'm trying to add a transition to the arrow icon, am I doing it right?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,226 @@
...
+  mounted () {
+    this.setInitialState(this.initialState)
+
+    if (this.categories) {
+      const setCategoriesNoExpanded = (category) => {
+        _.map(category[this.keyForChildren], innerCategory => {
+          setCategoriesNoExpanded(innerCategory)
+        })
+
+        return _.assign({}, category, { isExpanded: false })
```
fofi, line 160: Do you think that we could not define `isExpanded` by default? I mean, as we are setting it to a negative value, the falsy will remain the same

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+      this.filteredCategories = _.map(this.categories, setCategoriesNoExpanded)
```
fofi, line 163: This is actually the same as doing `this.filteredCategories = this.categories`, right?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+        <geo-tree-item
+          v-for="category in sortedCategories"
+          :key="category[keyForId]"
+          :category="category"
+          :key-for-id="keyForId"
+          :key-for-label="keyForLabel"
+          :checked-items="checkedItems"
+          :key-for-children="keyForChildren"
+          :description-icon="descriptionIcon"
+          @click="onCategoryClick"
```
Navas23, line 33: Try to unify namings here, call both using "on" or "handle"

src/elements/GeoTree/GeoTree.vue
Navas23: No idea! Keep it like this then

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+      default: false
+    },
+    /**
+     * Initial selected items
+    */
+    initialState: {
+      type: Object,
+      required: false
+    },
+    descriptionIcon: {
```
Navas23, line 130: Even the purpose is obvious, add the description here too please

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+    noResultsFound () {
+      return this.searchQuery && !this.filteredCategories.length
+    }
+  },
+  mounted () {
+    this.setInitialState(this.initialState)
+
+    if (this.categories) {
+      const setCategoriesNoExpanded = (category) => {
+        _.map(category[this.keyForChildren], innerCategory => {
```
Navas23, line 156: why using map and not forEach here?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+    },
+    noResultsFound () {
+      return this.searchQuery && !this.filteredCategories.length
+    }
+  },
+  mounted () {
+    this.setInitialState(this.initialState)
+
+    if (this.categories) {
+      const setCategoriesNoExpanded = (category) => {
```
Navas23, line 155: Is there a reason you did not create this function inside methods?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+      this.filteredCategories = _.map(this.categories, setCategoriesNoExpanded)
+    }
+  },
+  methods: {
+    handleSearching () {
+      this.$emit('search', this.searchQuery)
+
+      const searchByQuery = (category, query) => _.includes(category[this.keyForLabel], query)
+
+      const getFilteredCategories = (categories, query) => categories.reduce((carry, category) => {
```
Navas23, line 172: Could this be an independent method?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+          : carry
+      }, [])
+
+      this.filteredCategories = this.searchQuery
+        ? getFilteredCategories(this.categories, this.searchQuery)
+        : this.categories
+    },
+    onCategoryClick (clickedCategory) {
+      const isExpanded = (category) => category[this.keyForId] === clickedCategory[this.keyForId] ? !category.isExpanded : category.isExpanded
+      const toggleCategoriesExpanded = (categories) => _.map(categories, category => {
```
Navas23, line 203: Same here about independent methods

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,231 @@
...
+        )
+      })
+      this.filteredCategories = toggleCategoriesExpanded(this.filteredCategories)
+
+      this.$emit('click', clickedCategory)
+    },
+    handleCheckItem (categoryId, isChecked) {
+      this.checkedItems = isChecked
+        ? _.assign({}, this.checkedItems, { [categoryId]: true })
+        : _.omit(this.checkedItems, categoryId)
```
Navas23, line 222: Why do we omit it instead of setting it to false?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+    >
+      <geo-tree-item
+        v-for="categoryChildren in category[keyForChildren]"
+        :key="categoryChildren[keyForId]"
+        :category="categoryChildren"
+        :key-for-id="keyForId"
+        :key-for-label="keyForLabel"
+        :key-for-children="keyForChildren"
+        :checked-items="checkedItems"
+        @check="check"
```
Navas23, line 65: why not handleCheck?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+    },
+    /**
+     * Key to access to the children items of the category
+     */
+    keyForChildren: {
+      type: String,
+      required: false,
+      default: 'children'
+    },
+    checkedItems: {
```
Navas23, line 109: Add descriptions here please

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+    },
+    isIndeterminate (category) {
+      const isSomeChildSelected = category => {
+        return _.some(category[this.keyForChildren], subCategory => this.isChecked(subCategory))
+      }
+
+      return this.isChecked(category) ? false : isSomeChildSelected(this.category)
+    },
+    getTotalCategoryChildren (category) {
+      const sumOfChildren = category => !_.isEmpty(category[this.keyForChildren])
```
fofi, line 144: We could actually always use the same line
```
_.size(category[this.keyForChildren]) + _.sumBy(category[this.keyForChildren], sumOfChildren)
```

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+        ? category[this.keyForChildren].length + _.sumBy(category[this.keyForChildren], category => sumOfChildren(category))
+        : 0
+
+      return sumOfChildren(category)
+    },
+    getIfCategoryHasChildren (category) {
```
Navas23, line 150: Maybe we could simply call this "hasChildrens"

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,194 @@
...
+  methods: {
+    isChecked (category) {
+      const allAreChildrenSelected = category => category[this.keyForChildren] && _.every(category[this.keyForChildren], subCategory => this.isChecked(subCategory))
+
+      return this.getIfCategoryHasChildren(category)
+        ? allAreChildrenSelected(category)
+        : _.has(this.checkedItems, category[this.keyForId])
+    },
+    isIndeterminate (category) {
+      const isSomeChildSelected = category => {
```
Navas23, line 137: Same doubts than previous file, why not creating this helpers functions under methods?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+        return _.some(category[this.keyForChildren], subCategory => this.isChecked(subCategory))
+      }
+
+      return this.isChecked(category) ? false : isSomeChildSelected(this.category)
+    },
+    getTotalCategoryChildren (category) {
```
fofi, line 143: This could be a computed property, so we dont need to execute the method to calculate it every time

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+      const sumOfChildren = category => !_.isEmpty(category[this.keyForChildren])
+        ? category[this.keyForChildren].length + _.sumBy(category[this.keyForChildren], category => sumOfChildren(category))
+        : 0
+
+      return sumOfChildren(category)
+    },
+    getIfCategoryHasChildren (category) {
```
fofi, line 150: This could also be a computed

src/styles/elements/GeoTree/_geo-tree-item.scss
Navas23: I think you need to add here transform: rotate(0deg); and in the other one transition: transform 0.5s

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+  },
+  methods: {
+    isChecked (category) {
+      const allAreChildrenSelected = category => category[this.keyForChildren] && _.every(category[this.keyForChildren], subCategory => this.isChecked(subCategory))
+
+      return this.getIfCategoryHasChildren(category)
+        ? allAreChildrenSelected(category)
+        : _.has(this.checkedItems, category[this.keyForId])
+    },
+    isIndeterminate (category) {
```
fofi, line 136: This could also be a computed, right?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+      showActionButton: false
+    }
+  },
+  computed: {
+    categoryIcon () {
+      return this.getIfCategoryHasChildren(this.category) ? ['fal', 'chevron-right'] : null
+    }
+  },
+  methods: {
+    isChecked (category) {
```
fofi, line 129: Thsi could be a computed also

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+      })
+
+      this.$emit('check', category[this.keyForId], isChecked)
+    },
+    /**
+     * To manage the check action
+     * This method is listening to children check event
+     */
+    check (category, isChecked) {
+      if (this.getIfCategoryHasChildren(category)) {
```
fofi, line 178: We dont really need thus condition, right? `checkAll` is already doing the same as this method

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+        return this.checkAll(category, isChecked)
+      }
+
+      this.$emit('check', category[this.keyForId], isChecked)
+    },
+    /**
+     * The action button can't appear on parent categories
+     */
+    onMouseOver (category) {
+      if (!this.getIfCategoryHasChildren(category)) {
```
fofi, line 188: I am not sure if we want this condition, maybe we want to display something on the parents, right?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,191 @@
...
+      <geo-tree-item
+        v-for="categoryChildren in category[keyForChildren]"
+        :key="categoryChildren[keyForId]"
+        :category="categoryChildren"
+        :key-for-id="keyForId"
+        :key-for-label="keyForLabel"
+        :key-for-children="keyForChildren"
+        :checked-items="checkedItems"
+        @check="check"
+        @click="handleClick"
```
fofi, line 66: Why do we need the click if we have the check?

fofi (commented):
First round

src/elements/GeoTree/GeoTree.vue
asanzdj: I prefer to use defaults to avoid errors, but I can try if it's working without it

src/styles/elements/GeoTree/_geo-tree-item.scss
asanzdj: Okey, I'm going to try that, thank you

src/elements/GeoTree/GeoTree.vue
asanzdj: I use on when the component is not handling the action, it emits and using handling when the component is who handle it.

src/elements/GeoTree/GeoTree.vue
asanzdj: Yes, this is a pending issue (add all the descriptions)

src/elements/GeoTree/GeoTree.vue
asanzdj: Because I'm returning a value to get a new array

src/elements/GeoTree/GeoTree.vue
asanzdj: I'm using it only inside this method

src/elements/GeoTree/GeoTree.vue
asanzdj: Yes, but why? (This method is WIP because maybe it disapears from here)

src/elements/GeoTree/GeoTree.vue
Navas23: But... inside function "onCategoryClick" you are handling the click, right?

src/elements/GeoTree/GeoTree.vue
asanzdj: Here I don't have the famous method getThrutyKeys, so I decided to extract it from the selected variable

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Yes, that is wrong

src/elements/GeoTree/GeoTree.vue
Navas23: I still think it's better to keep it outside (just for being verbose), but not strong opinion. What do you think @fofi ?

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Yes, it's in my todo list

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Okey, I would want to be specific

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Because there are only used here

src/elements/GeoTree/GeoTree.vue
Navas23: But you are not using the new array (you are not storing it)

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,226 @@
...
+    /**
+     * Initial selected items
+    */
+    initialState: {
+      type: Object,
+      required: false
+    },
+    descriptionIcon: {
+      type: Array,
+      default: () => []
```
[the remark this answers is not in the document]
asanzdj: To prevent it to break, but I'm going to try it without default

src/elements/GeoTree/GeoTree.vue
Navas23: As I told you, just for readibility of the code... but not sure

src/elements/GeoTree/GeoTree.vue
asanzdj: Which variable? I think this logic is related with the component, so it should to manage it. Some things are emitted out

src/elements/GeoTree/GeoTree.vue
asanzdj: Okey, I'm going to extract it into a prop

src/styles/elements/GeoTree/_geo-tree-item.scss
Navas23: Let me know if it does not work, CSS gets tricky sometimes

src/elements/GeoTree/GeoTree.vue
asanzdj: No, I'm adding to categories the isExpand property

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Okey, I will see it

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: I have to investigate about it because I was having problems using computed

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: I will take a look

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: That's a condition specified in the task that was validated but I'll ask again

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: They're different things, check is used to click on the checkbox, click is used to click on the list item

src/elements/GeoTree/GeoTree.vue
asanzdj: Yes, I'm stupid

src/elements/GeoTree/GeoTree.vue
asanzdj: Ah, okey, now I'm getting the point

package.json
```diff
@@ -188,5 +188,8 @@
     "not BlackBerry > 0",
     "not OperaMini all",
     "not OperaMobile > 0"
-  ]
+  ],
+  "dependencies": {
+    "fuzzaldrin-plus": "^0.6.0"
```
fofi, line 193: We have the rest of dependencies as peerDependencies, we should probably keep that the same

src/elements/GeoTree/GeoTree.vue
fofi: I think we will have more flexibility if the state is not inside the component, this is because for example if we want to init this component with some items checked, we wont be able to do it.

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,245 @@
...
+     * Icon used to alert about some extra info displayed in a popover
+     */
+    descriptionIcon: {
+      type: Array,
+      required: false
+    },
+    /**
+     * Should appear all items collapsed or not initially
+     */
+    isInitiallyExpanded: {
```
fofi, line 141: I am not sure if we will want to apply this to every children, probably just to some of them.

I think we will want to just open the first level for example, but the rest we will want to keep them closed

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,245 @@
...
+      return this.searchQuery && !this.filteredCategories.length
+    }
+  },
+  mounted () {
+    this.setInitialState(this.initialState)
+
+    if (this.categories) {
+      const setCategoriesNoExpanded = (category) => {
+        _.forEach(category[this.keyForChildren], innerCategory => {
+          setCategoriesNoExpanded(innerCategory)
```
fofi, line 169: This will actually return a new object with `isExpanded` that we are not using

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+import _ from 'lodash'
+
+export default {
+  name: 'GeoTreeItem',
+  status: 'ready',
+  release: '29.9.0',
+  props: {
+    category: {
+      type: Object,
+      required: false,
```
fofi, line 85: I think this should be required, right?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      return this.isChecked ? false : isSomeChildSelected(this.category)
+    },
+    isChecked () {
+      const allAreChildrenSelected = category => category[this.keyForChildren] && _.every(category[this.keyForChildren], subCategory => !!this.checkedItems[subCategory[this.keyForId]])
+
+      return this.hasChildren
+        ? allAreChildrenSelected(this.category)
+        : _.has(this.checkedItems, this.category[this.keyForId])
```
fofi, line 148: Why using has? shouldn't it be that `this.checkedItems[this.category[this.keyForId]]` is truthy?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+      return this.totalCategoryChildren > 0
+    }
+  },
+  methods: {
+    /**
+     * On list item click
+     */
+    handleClick (category) {
+      if (!this.hasChildren) {
+        this.$emit('check', category, !this.isChecked)
```
fofi, line 165: If the "click" means " click is used to click on the list item", why are we emitting "check" ?

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,245 @@
...
+      isLoading: false,
+      checkedItems: {},
+      filteredCategories: []
+    }
+  },
+  computed: {
+    sortedCategories () {
+      return _.sortBy(this.filteredCategories, [this.keyForLabel])
+    },
+    noResultsFound () {
```
fofi, line 159: Usually is more readable to use positive variables, like `hasResults`

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,245 @@
...
+  },
+  methods: {
+    handleSearching () {
+      this.$emit('search', this.searchQuery)
+
+      const clearString = string => _.toLower(_.deburr(string))
+
+      const getFilteredCategories = (categories, query) => categories.reduce((carry, category) => {
+        const isCategoryMatching = fuzzAldrin.score(category[this.keyForLabel], query) > 0
+        const categories = category[this.keyForChildren] ? getFilteredCategories(category[this.keyForChildren], query) : null
```
Navas23, line 186: Probably calling this const something like "matchedChildrenCategories" could make this more verbose

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,245 @@
...
+
+        if (categories && categories.length) {
```
Navas23, line 188: And what do you think here about using a ternary operator? Do you think will make this more or less verbose?

src/elements/GeoTree/GeoTree.vue
asanzdj: Mmm, you're right, so I'm going to change it to receive an array of category ids to expand

src/elements/GeoTree/GeoTree.vue
asanzdj: Okey, I'm going to refactor this using new idea

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Yes, I thought it like: maybe you're getting data from async call and at first moment you haven't got the data but that isn't the problem of this component

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: :flushed:

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: Mmm, maybe I have  to change it to call the checkAll method. That's because if user clicks on a category that haven't got children it should mark it as check

src/elements/GeoTree/GeoTree.vue
asanzdj: Oki

src/elements/GeoTree/GeoTree.vue
asanzdj: Oki

src/elements/GeoTree/GeoTree.vue
asanzdj: Okey, In my opinion I don't like very much have something like this, but I can implement it here because they are easy conditions

```js
return check 
? positive_result
: check2 ? positive_result_2 : negative_result
```

Navas23 (approved):

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,244 @@
...
+import fuzzAldrin from 'fuzzaldrin-plus'
+
+export default {
+  name: 'GeoTree',
+  status: 'ready',
+  release: '29.9.0',
+  props: {
+    /**
+     * Text to display as placeholder of the search input
+   */
```
fofi, line 56: Don't forget to indent properly all the comments here

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,244 @@
...
+              }
+            )
+          ]
+          : isCategoryMatching
+            ? [
+              ...carry,
+              _.assign(category, { matches: fuzzAldrin.match(clearString(category[this.keyForLabel]), clearString(query)) })
+            ]
+            : carry
+      }, [])
```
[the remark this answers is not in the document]
fofi: I think something like this will be a little bit more readable, or at least less code
`
const matchingCat = isCategoryMatching
 ? _.assign(category, { matches: fuzzAldrin.match(clearString(category[this.keyForLabel])
 : null

if (_.size(matchedChildrenCategories)) {
matchingCat.isExpanded = true
matchingCat[this.keyForChildren] = matchedChildrenCategories
}

return _.concat(carry, _.filter([matchingCat]))
`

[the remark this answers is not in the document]
asanzdj: Okey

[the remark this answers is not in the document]
asanzdj: What do you think now? I extracted it in little blocks

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,244 @@
...
+        const basicCategory = _.assign(
+          {},
+          category,
+          { matches: fuzzAldrin.match(clearString(category[this.keyForLabel]), clearString(query)) })
+        const matchedSubcategories = category[this.keyForChildren]
+          ? getFilteredCategories(category[this.keyForChildren], query)
+          : null
+
+        /*
+        * If some subcategory match with the searched text it should display complete category tree
```
fofi, line 181: Is this true? Reading the code I understand that we are displaying the matching subcategories, not the complete tree

src/elements/GeoTree/GeoTree.vue
```diff
@@ -0,0 +1,244 @@
...
+                isExpanded: true
+              }
+            )
+          ]
+        }
+
+        /*
+        * If some category match with the searched text it should display it
+        */
+        if (isCategoryMatching) {
```
fofi, line 199: Why do we need this condition?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+    handleClick (category) {
+      if (!this.hasChildren) {
+        this.handleCheckAll(category, !this.isChecked)
+      }
+      this.$emit('click', category)
+    },
+    /**
+    * To check all items of a category
+    */
+    handleCheckAll (category, isChecked) {
```
fofi, line 173: Why handleCheckAll and not just handleCheck?

src/elements/GeoTree/GeoTreeItem.vue
fofi: I think in native html if you click on a label that is assigned to a checkbox, it will always check the checkbox.

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+  },
+  methods: {
+    /**
+    * On list item click
+    */
+    handleClick (category) {
+      if (!this.hasChildren) {
+        this.handleCheckAll(category, !this.isChecked)
+      }
+      this.$emit('click', category)
```
fofi, line 168: Why do we want to emit the click also if we have clicked on a element without children?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+    }
+  },
+  data () {
+    return {
+      showActionButton: false
+    }
+  },
+  computed: {
+    categoryIcon () {
+      return this.hasChildren ? ['fal', 'chevron-right'] : null
```
fofi, line 135: Do we want to customize also this icon?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -0,0 +1,184 @@
...
+    * Icon used to alert about some extra info displayed in a popover
+    */
+    descriptionIcon: {
+      type: Array,
+      default: () => []
+    }
+  },
+  data () {
+    return {
+      showActionButton: false
```
fofi, line 130: Is a good practice for boolean to use names like `isActionButtonVisible` as they are more descriptive, `showActionButton` could be a method to actually make the button visible

test/unit/specs/elements/GeoTree/GeoTree.spec.js
```diff
@@ -0,0 +1,196 @@
...
+            id: 'garlic',
+            label: 'Garlic'
+          }
+        ]
+      }
+    ]
+  }
+]
+
+describe.only('GeoTree', () => {
```
fofi, line 122: Do nto forget to remove the .only

test/unit/specs/elements/GeoTree/__snapshots__/GeoTree.spec.js.snap
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
fofi, line 1: Why do we have this file?

fofi (commented):
Nice job with the new tests

test/unit/specs/elements/GeoTree/GeoTree.spec.js
asanzdj: I'm still working on the test

test/unit/specs/elements/GeoTree/__snapshots__/GeoTree.spec.js.snap
asanzdj: Because it's a way to test that a component is working as previous version in an easy way. https://www.digitalocean.com/community/tutorials/vuejs-jest-snapshot-testing-in-vue (If you update your component you must update the test and the snapshot too)

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: I don't think so but I can move it into a prop.

src/elements/GeoTree/GeoTree.vue
asanzdj: The idea is show the matching categories but opening the parent categories

src/elements/GeoTree/GeoTree.vue
asanzdj: You don't want to add a category that is not matching with the searched text

src/elements/GeoTree/GeoTreeItem.vue
asanzdj: If you click on:

- Element with subcategories: it should expand the category
- Element without subcategories: it should check/uncheck the category

src/elements/GeoTree/GeoTreeItem.vue
fofi: Lets do it just if needed

src/elements/GeoTree/GeoTree.vue
```diff
@@ -142,102 +135,86 @@ export default {
...
-      return _.sortBy(this.filteredCategories, [this.keyForLabel])
-    },
     hasResults () {
-      return !this.searchQuery || (this.searchQuery && this.filteredCategories.length)
+      return !this.searchQuery || (this.searchQuery && this.categoriesToShow.length)
+    }
+  },
+  watch: {
+    categories (newCategories) {
+      this.sortCategories(newCategories)
```
fofi, line 161: Why a watch and not a computed? I think we can have a computed called `sortedCategories`, and use it everywhere instead of `categories`

src/elements/GeoTree/GeoTree.vue
```diff
@@ -142,102 +135,86 @@ export default {
...
     }
   },
   mounted () {
-    this.setInitialState(this.initialState)
-    this.filteredCategories = this.categories
+    this.sortCategories(this.categories)
```
fofi, line 165: If we have it as a computed we dont need this also

src/elements/GeoTree/GeoTree.vue
```diff
@@ -142,102 +135,86 @@ export default {
...
       }, [])
 
-      this.filteredCategories = this.searchQuery
-        ? getFilteredCategories(this.categories, this.searchQuery)
+      function clearString (string) {
+        return _.toLower(_.deburr(string))
+      }
+    },
+    handleSearching () {
+      this.categoriesToShow = this.searchQuery
```
fofi, line 194: Why the rename? I think `filteredCategories` was super clear, maybe more than `categoriesToShow`

src/elements/GeoTree/GeoTree.vue
```diff
@@ -142,102 +135,86 @@ export default {
...
-          {
-            isExpanded: isExpanded(category),
-            [this.keyForChildren]: children
-          }
-        )
-      })
-      this.filteredCategories = toggleCategoriesExpanded(this.filteredCategories)
+      clickedCategory.isExpanded = !clickedCategory.isExpanded
 
       this.$emit('click', clickedCategory)
```
fofi, line 201: Why do we want to emit this click?

test/unit/specs/elements/GeoTree/GeoTree.spec.js
```diff
@@ -152,45 +162,316 @@ describe.only('GeoTree', () => {
...
+
+    expect(wrapper.find('.geo-input').exists()).toBe(true)
+    wrapper.find('.geo-input input').setValue('banana')
+
+    expect(wrapper.vm.filteredCategories).toEqual(expectedFilteredCategories)
   })
 
-  it('should filter the categories when on user searching', () => {
+  // TODO: Poner bien este texto, crear un describe solo para el buscador.
+  it('hijo + padre', () => {
```
fofi, line 201: I guess this is a WIP. but lets find a good name for the tests

src/elements/GeoTree/GeoTree.vue
asanzdj: I was thinking: maybe outside needs to know when a category is click, but thinking it now it want to know when a category is checked not clicked

src/elements/GeoTree/GeoTree.vue
```diff
@@ -194,7 +197,7 @@ export default {
       }
     },
     handleSearching () {
-      this.categoriesToShow = this.searchQuery
+      this.filteredCategories = this.searchQuery
         ? this.filterCategories(this.categories, this.searchQuery)
```
fofi, line 201: Remember we have to filter `sortedCategories` in order to keep the sort

fofi: We should just use `this.categories` to create `sortedCategories`

test/unit/specs/elements/GeoTree/__snapshots__/GeoTree.spec.js.snap
fofi: I don't really like snapshot tests, in my opinion they aren't really useful as we have to change them every time we change anything

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
```diff
@@ -0,0 +1,191 @@
...
+    expect(wrapper.find('.geo-tree-item__total-items').exists()).toBe(true)
+    expect(wrapper.find('.geo-tree-item__total-items').text()).toBe(`(${5})`)
+  })
+
+  it('should render a list of subcategories on category click (it it has subcategories)', async () => {
+    const wrapper = getWrapper()
+
+    wrapper.find('.geo-tree-item').trigger('click')
+    await Vue.nextTick()
+    await wrapper.vm.$forceUpdate()
```
asanzdj, line 77: I don't think this is the solution but I'm having a lot of problems because the instance isn't be refreshed

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
```diff
@@ -0,0 +1,191 @@
...
+    expect(wrapper.find('#avocado').element.checked).toBe(false)
+
+    wrapper.find('#avocado').setChecked()
+    expect(wrapper.find('#avocado').element.checked).toBe(true)
+
+    wrapper.find('#avocado').setChecked(false)
+    expect(wrapper.find('#avocado').element.checked).toBe(false)
+  })
+
+  it('should mark as checked and then as unchecked a category and all of its children nodes', async () => {
```
asanzdj, line 138: This test is not working

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
```diff
@@ -0,0 +1,184 @@
...
+
+describe('GeoTreeItem', () => {
+  const getWrapper = (props = {}) => shallowMount(GeoTreeItem, {
+    propsData: _.assign(
+      {},
+      {
+        keyForLabel: 'label',
+        keyForSubcategory: 'subcategories',
+        keyForId: 'id',
+        category: props.category || CATEGORY,
```
Navas23, line 44: I think this or is not needed here, you could just simply say category: CATEGORY

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
Navas23: Using this works?

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
Navas23: In which step stops working?

test/unit/specs/elements/GeoTree/GeoTree.spec.js
```diff
@@ -0,0 +1,546 @@
...
+                label: 'Pepper',
+                matches: []
+              }
+            ]
+          }
+        ]
+      }
+    ]
+
+    expect(wrapper.vm.filteredCategories).toEqual(expectedSortedCategories)
```
fofi, line 278: We could check also sortedCategories

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
fofi: Do we need both?

fofi: I don't understand why we need this, isn't this supposed to test async stuff?

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
fofi: I think the problem is that we are not updating `checkedItems`

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
asanzdj: It wasn't listening the changes in the vue instance (html + internal data), so I had to mixed the solutions that people provide

fofi (approved):

test/unit/specs/elements/GeoTree/GeoTree.spec.js
asanzdj: Okey

fofi (approved):

fofi (approved):

src/elements/GeoTree/GeoTree.vue
```diff
@@ -200,6 +211,11 @@ export default {
             })
             : category
         ), this.keyForLabel)
+    },
+    handleToggleExpand (clickedCategory) {
+      const isExpanded = !!this.expandedCategories[clickedCategory[this.keyForId]]
+
+      this.expandedCategories = _.assign({}, this.expandedCategories, { [clickedCategory[this.keyForId]]: !isExpanded })
```
fofi, line 218: Why are we creating a new object every time we toggle a category?

fofi: Maybe with $set is enoguh

PR #76, opened 2020-10-29
(fix) Removed outdated prop in geo-button

asanzdj:
Jira task:

https://geoblink.atlassian.net/browse/CORE-8496

### DIFF
--- a/src/elements/GeoButton/GeoButton.template.html
+++ b/src/elements/GeoButton/GeoButton.template.html
@@ -1,7 +1,6 @@
 <geo-button
   v-bind="$props"
   :type="type"
-  :isPrimary="isPrimary"
   @click="$emit('click', $event)"
 >
   <slot />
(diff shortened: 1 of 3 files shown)


src/elements/GeoButton/GeoButton.template.html
```diff
@@ -1,12 +1,11 @@
...
   @click="$emit('click', $event)"
 >
   <slot />
   <slot
     slot="loading"
     name="loading"
   />
-</geo-button>
\ No newline at end of file
+</geo-button>
```
asanzdj, line 11: I don't know why this appear like a change, my code editor is adding a new line at the end but I can't commit it.

Navas23 (commented):
We do not use this prop in the app right? do we have some functionality related to this prop?

fofi (approved):

asanzdj:
> We do not use this prop in the app right? do we have some functionality related to this prop?

No, now you have to specify the button type (the prop name is `type`), this prop was never declared but used in the template.

Navas23 (approved):

PR #77, opened 2020-10-30
Feature/core 8481 not modify input text in geoborderedboxheadersearcher

asanzdj:
Jira Task:
https://geoblink.atlassian.net/browse/CORE-8481

### DIFF
--- a/src/elements/GeoBorderedBox/GeoBorderedBoxHeaderSearchForm.vue
+++ b/src/elements/GeoBorderedBox/GeoBorderedBoxHeaderSearchForm.vue
@@ -94,7 +94,7 @@ export default {
          * @event input
          * @type {string}
          */
-        this.$emit('input', _.deburr(value))
+        this.$emit('input', value)
       })
     }
   },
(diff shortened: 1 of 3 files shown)


src/elements/GeoBorderedBox/GeoBorderedBoxHeaderSearchForm.vue
```diff
@@ -94,7 +94,7 @@ export default {
          * @event input
          * @type {string}
          */
-        this.$emit('input', _.deburr(value))
+        this.$emit('input', value)
```
fofi, line 97: We should apply this now on the search for the tree element, right?

asanzdj: In that component I added a string cleaner function to avoid this problem, so it isn't needed. You can check it in `src/elements/GeoTree/GeoTree.vue`, L.190 and L.204

fofi (approved):

Navas23 (approved):

PR #78, opened 2020-11-03
Feature/core 8451 geo list item emiting events when is disabled

asanzdj:
**Jira task:**

https://geoblink.atlassian.net/browse/CORE-8451

**Related info to solve the task:**

- Vue $listeners: https://vuejs.org/v2/api/#vm-listeners
- Combine $listeners with control events: https://github.com/vuejs/vue/issues/7042
- Transparent components ($listeners + custom events): https://zendev.com/2018/05/31/transparent-wrapper-components-in-vue.html
- Functional components: https://vue-loader.vuejs.org/guide/functional.html, https://v3.vuejs.org/guide/migration/functional-components.html


src/elements/GeoList/GeoListItem.vue
```diff
@@ -1,22 +1,22 @@
-<template functional>
+<template>
```
Navas23, line 1: Why did you remove the functional?

src/elements/GeoList/GeoListItem.vue
```diff
@@ -1,22 +1,22 @@
...
   <component
-    :is="props.wrapperTag"
-    :ref="data.ref"
-    v-bind="data.attrs"
-    :class="[data.class, data.staticClass, {
+    :is="wrapperTag"
+    :ref="$attrs.ref"
```
Navas23, line 4: This is because it's not functional, right?

src/elements/GeoSegmentedControl/GeoInfoSegmentedControlItem.vue
```diff
@@ -28,9 +29,36 @@ export default {
...
+      default: false
+    },
+
+    /**
+     * When is true the component will display an active style.
+     */
+    active: {
+      type: Boolean,
+      default: false
+    }
```
fofi, line 61: Why are we adding these props if they are already in the mixin?

src/elements/GeoSegmentedControl/GeoInfoSegmentedControlItem.vue
```diff
@@ -1,16 +1,16 @@
-<template functional>
+<template>
```
fofi, line 1: We had the code like this because this was a functional component, probably now that is not anymore, we can modify it like a normal component

fofi (commented):
Should we add this to other components?

src/elements/GeoList/GeoListItem.vue
asanzdj: If it's functional I can't take control about its events.

src/elements/GeoList/GeoListItem.vue
asanzdj: Yes

src/elements/GeoSegmentedControl/GeoInfoSegmentedControlItem.vue
asanzdj: Ah, okey, I'll try it

src/elements/GeoSegmentedControl/GeoInfoSegmentedControlItem.vue
asanzdj: Because I didn't notice about that but I've already removed then

src/elements/GeoList/GeoListItem.vue
```diff
@@ -1,22 +1,21 @@
-<template functional>
+<template>
   <component
-    :is="props.wrapperTag"
-    :ref="data.ref"
```
Navas23, line 4: We don't need this ref now?

test/unit/specs/elements/GeoList/GeoListItem.spec.js
```diff
@@ -1,19 +1,18 @@
...
+import { shallowMount } from '@vue/test-utils'
 import GeoListItem from '@/elements/GeoList/GeoListItem.vue'
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import { library } from '@fortawesome/fontawesome-svg-core'
 import { fas } from '@fortawesome/free-solid-svg-icons'
 
 library.add(fas)
 
-describe('GeoListItem', () => {
+describe.only('GeoListItem', () => {
```
Navas23, line 9: Why do we need the only here?

test/unit/specs/elements/GeoList/GeoListItem.spec.js
```diff
@@ -22,42 +21,8 @@ describe('GeoListItem', () => {
...
-  it('Should render with the provided static class', () => {
-    const wrapper = mount(GeoListItem, {
-      context: {
-        staticClass: 'my-static-class'
-      }
-    })
-    expect(wrapper.find('div.geo-list-item.my-static-class').exists()).toBe(true)
-  })
-
-  it('Should render with the provided dynamic class', () => {
```
Navas23, line 34: Why did you remove so many tests?

src/elements/GeoList/GeoListItem.vue
asanzdj: I did some tries to know if passing a ref is necessary or not. And I've noticed that is not needed.
For example, here, in the live example I have that ref without receiving it as a prop in the component.

Screenshot 2020-11-03 at 12 38 42

test/unit/specs/elements/GeoList/GeoListItem.spec.js
asanzdj: I'm fixing test and  I've forgotten about this

test/unit/specs/elements/GeoList/GeoListItem.spec.js
asanzdj: This tests was associated to a weird behaviour, dynamic and static classes. I removed that behaviour so I removed related tests too. I noticed that you can add extra classes to the component instance to style it and it isn't needed to receive them as a prop by the component.

Navas23 (approved):

test/unit/specs/elements/GeoList/GeoListItem.spec.js
```diff
@@ -104,22 +67,20 @@ describe('GeoListItem', () => {
 
   it('Should emit click event', function () {
```
fofi, line 68: We should create a test testing the disabled not emitting, right?

test/unit/specs/elements/GeoSegmentedControl/GeoSegmentedControlItem.spec.js
```diff
@@ -30,37 +30,37 @@ describe('GeoSegmentedControlItem', () => {
...
-        expect(wrapper.emitted().click).toBeTruthy()
-        done()
-      } catch (error) {
-        done(error)
-      }
-    })
+    expect(clickListener).toHaveBeenCalled()
   })
 
   it('Should not emit an event when it\'s disabled', function () {
```
fofi, line 47: Why Was this test working perfectly?

test/unit/specs/elements/GeoList/GeoListItem.spec.js
asanzdj: Added

fofi (approved):

PR #80, opened 2020-11-11
Feature/core 8525 allow date as quadrant threshold

fofi:


Navas23 (commented):
We don't need to do any changes in the functionality?

fofi:
> We don't need to do any changes in the functionality?

No, the functionality was already working, but we were receiving some errors

Navas23 (approved):
Aprobadito miarma, pero no te olvides de quitar el tag beta al desplegar

PR #82, opened 2020-12-23
Feature/core 8577 update geo value

Navas23:
Geo-value updates to support KPIs with benchmarks


dletran (approved):
Looks good to me

fofi (approved):
LGTM

PR #84, opened 2021-02-03
Feature/core 8609 replace text element with html to allow text ellipsis with css

dletran:


src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
```diff
@@ -422,3 +427,46 @@ export function createCustomizedTickArray (firstOfDomain, lastOfDomain, tickCoun
...
+      div.attr('class', 'geo-trimmed-content geo-chart-axis-text--right')
+      break
+    }
+    default:
+      // we keep the top and bottom axis as it is
+      break
+  }
+
+  function createForeignElem () {
+    text.attr('style', 'visibility: hidden')
```
Navas23, line 463: Why use visibility instead of display? It will keep the space using visibility (so it will be proccessed)

Navas23 (commented):
Looks good, some questions:
- Have we tried in the whole app that every chart work as before?
- What happen if we change geo-trimmed component? It will imply do modifications here right?

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
dletran: I used visibility so it's rendered and we can retrieve the `x` attribute value from it, as we're using coordinates to place the foreignObject in the svg group the space the text element takes is not affecting it

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
```diff
@@ -422,3 +427,46 @@ export function createCustomizedTickArray (firstOfDomain, lastOfDomain, tickCoun
   })
   return tickToDisplay
 }
+
+function positionTickHtmlLabel (group, text, content, singleAxisOptions, globalAxesConfig) {
+  let foreignElem = group.select('foreignObject')
```
fofi, line 432: Can we use const here?

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
fofi, line 463: IMHO is kinda unexpected to hide something inside a method called `createForeignElem`

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
fofi: I think we could just call here `createForeignElem` and this one will get the existing one or just create a new one.

Maybe we should also rename that method

fofi (commented):
Apart from the comments this looks really good!

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
Navas23: Nice

dletran: yep, I'll move it outside

src/elements/GeoChart/GeoChartAxis/GeoChartAxis.js
dletran: Hmmm ok, I can rename it to getForeignElem and move the logic to retrieve it or create it if it does not exists inside, we can't create a new one if it already exists because then we'd be adding new elements on every redraw

test/unit/specs/elements/GeoChart/GeoChartAxis.spec.js
```diff
@@ -1025,5 +1025,108 @@ describe('GeoChartAxis', function () {
...
+              }
+            }
+          })
+
+          flushD3Transitions()
+
+          expect(wrapper.find('.geo-chart').exists()).toBe(true)
+          expect(wrapper.find('.geo-chart-axis').exists()).toBe(true)
+          expect(wrapper.find('.geo-chart-axis .tick').exists()).toBe(true)
+          expect(wrapper.find('.geo-chart-axis .tick text').exists()).toBe(true)
```
fofi, line 1098: I think we should check the visibility

test/unit/specs/elements/GeoChart/GeoChartAxis.spec.js
```diff
@@ -1025,5 +1025,108 @@ describe('GeoChartAxis', function () {
...
+          expect(wrapper.find('.geo-chart-axis .tick foreignObject').exists()).toBe(true)
+
+          const allTicks = wrapper.findAll('.geo-chart-axis .tick')
+
+          for (let i = 0; i < allTicks.length; i++) {
+            const textContent = allTicks.at(i).find('text').text()
+            const htmlContent = allTicks.at(i).find('foreignObject div span').text()
```
fofi, line 1105: Probably we should also check the div class (left vs right)

fofi (approved):

test/unit/specs/elements/GeoChart/GeoChartAxis.spec.js
dletran: agreed

test/unit/specs/elements/GeoChart/GeoChartAxis.spec.js
dletran: agreed

Navas23 (approved):

dletran:
feature/core-8609 ranking chart store names sometimes get out of the box

PR #86, opened 2021-02-08
Feature/core 8609 increase axis space

dletran:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "29.9.15",
+  "version": "29.9.16-beta.1",
```
Navas23, line 3: Do we want this to be a beta?

dletran: nope, but I thought to wait for the approval before publishing and changing the version

Navas23 (approved):

fofi (approved):

PR #88, opened 2021-03-05
Feature/core 8672 geo pill variant

dletran:


src/styles/elements/GeoPill/_geo-pill.scss
```diff
@@ -36,3 +33,11 @@
   font-size: $font-size-small;
   height: $space-large;
 }
+
+.geo-pill--outline {
+  background: $color-white;
+  border: 1px solid $color-yellow-light;
+  color: $color-grey-darker;
+  font-size: $font-size-small;
```
Navas23, line 41: We don't need font-size and height here, right?

fofi (approved):

src/styles/elements/GeoPill/_geo-pill.scss
dletran: Yes, we want to replicate the 'opaque' style but having a border instead of a solid color background

Navas23: Sure, my bad

Navas23 (approved):

PR #89, opened 2021-04-09
Feature/core 8692 select toggle button

dletran:


fofi (approved):

test/unit/specs/elements/GeoDropdown/GeoDropdownRegularButton.spec.js
```diff
@@ -93,4 +94,31 @@ describe('GeoDropdownRegularButton', () => {
...
+      },
+      propsData: {
+        icon: ['fas', 'user']
+      }
+    })
+
+    expect(wrapper.find('.geo-dropdown__regular-button-container__icon--left').exists()).toBe(true)
+  })
+
+  it('Should apply proper class when icon position left', function () {
```
Navas23, line 111: This should be icon position right

Navas23 (changes requested):
Only one small comment, good job!

test/unit/specs/elements/GeoDropdown/GeoDropdownRegularButton.spec.js
dletran: ah true! let me change it

Navas23 (approved):

PR #92, opened 2021-06-04
Feature/core 8915 separated utils file

dletran:


build/webpack.system.conf.js
```diff
@@ -73,13 +83,10 @@ const plugins = [
   new MergeWebpackPlugin({
     files: {
       [utils.assetsSystemPath('system.utils.scss')]: [
-        './src/assets/tokens/tokens.scss',
-        './src/styles/_spacing.scss',
-        './src/styles/_fontsMaps.scss',
-        './src/styles/_mixins.scss',
-        './src/styles/_functions.scss',
+        ...scssToolsFiles,
```
fofi, line 86: Do we want to "duplicate" the tools in both files?

Navas23 (approved):

build/webpack.system.conf.js
```diff
@@ -73,13 +83,10 @@ const plugins = [
...
         ...elementsSCSSPaths
-      ]
+      ],
+      [utils.assetsSystemPath('tools/system.tools.scss')]: scssToolsFiles
```
fofi, line 89: Why "tools" ? I am not sure if the user can understand the content of the file

fofi (commented):
Apart from the comments, nice idea!

build/webpack.system.conf.js
dletran: Yes because that way we keep backwards compatibility for files importing the whole system.utils file that also use the variables, functions or mixins. Otherwise it would be a breaking change since it would force to import both

build/webpack.system.conf.js
dletran: I was going to use 'utils' but we're already using it for the generated scss file, maybe we can rename it 'tokens' to match the naming in the DS webpage wdyt?

fofi: I like it more, I think is more accurate

fofi (approved):

PR #93, opened 2021-06-21
Feature/core 8955 allow multiple weeks and quarters

Navas23:


fofi (approved):

src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -411,51 +399,115 @@ export default {
...
       this.emitFromDate({ fromDate: this.fromRawDate })
       this.emitToDate({ toDate: this.toRawDate })
+
+      this.resetInputFocus()
     },
 
     selectWeek ({ fromDate, toDate }) {
-      this.fromRawDate = fromDate
-      this.toRawDate = toDate
+      const computedDayForDifference = _.isNull(fromDate)
```
dletran, line 452: Looks like the logic for selectWeek, selectQuarter and selectYear is pretty similar, could we try to have a common function for them?

dletran (commented):
Apart from the comment 👍

src/elements/GeoCalendar/GeoCalendar.vue
Navas23: It is almost similar but they differ a bit, so I only extracted what it's exactly the same

dletran (approved):

PR #96, opened 2021-07-07
Feature/core 8816 throttle tooltip log

Navas23:


dletran (approved):

src/elements/GeoTooltip/GeoTooltip.vue
```diff
@@ -157,6 +157,12 @@ export default {
 
     onTriggerTargetMouseoverBound () {
       return this.onTriggerTargetMouseover.bind(this)
+    },
+
+    canNotFitWarnThrottled () {
+      return throttle(() => {
```
fofi, line 163: Why throttle and not debounce?

Navas23: Because we want to get the first warning instantly

fofi: In our thottle method, we are executing it with `leading: true` and `trailing: true`, so we will only get instant if its executed once, if not it will be executed at the end

fofi: But I guess is fine for our use case

fofi (approved):

PR #98, opened 2021-09-22
Feature/core 9155 fix latest date year

dletran:


fofi (approved):
Can this happen with other granularity?

dletran:
> Can this happen with other granularity?

@fofi I'll check it for month too, which probably it's the same, and I'll apply the same solution in that case

Navas23 (approved):

src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -435,12 +438,12 @@ export default {
...
         : {
-          start: unverifiedRange.end,
-          end: unverifiedRange.start
+          start: startOfQuarter(unverifiedRange.end),
+          end: endOfQuarter(unverifiedRange.start)
         }
 
-      this.fromRawDate = validatedRange.start && startOfQuarter(validatedRange.start)
-      this.toRawDate = validatedRange.end && endOfQuarter(validatedRange.end)
+      this.fromRawDate = validatedRange.start
```
fofi, line 445: Why changing this code?

src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -382,7 +384,7 @@ export default {
           this.isFromInputFieldExplicitlyFocused ||
           distanceToFromDate < distanceToToDate
 
-      const unverifiedRange = this.getUnverifiedRange(firstDayOfMonth, lastDayOfMonth, isSettingFromDate)
+      const unverifiedRange = this.getUnverifiedRange(firstDayOfMonth, possibleLastDayOfSelectedMonth, isSettingFromDate)
```
fofi, line 387: Should we do the same with the initial day?

fofi (changes requested):
I think we should add some tests for this cases

src/elements/GeoCalendar/GeoCalendar.vue
dletran: if the initial day is before the latest date we won't allow to select it in the picker itself

src/elements/GeoCalendar/GeoCalendar.vue
dletran: If we apply the endOfQuarter after the validatedRange we'd use the end of the quarter even if it's after the latest date

src/elements/GeoCalendar/GeoCalendar.vue
fofi: I mean if the firstDayOfMonth is smaller than earliestDate ?

dletran: ah good point! I'll have a look

fofi (commented):
What about selectWeek ?

dletran:
> What about selectWeek ?

@fofi we don't allow selecting a "non-complete" week on the picker, although we could add the verification just in case we change that in the future so everything works the same way, wdyt?

fofi (approved):

fofi:
> > What about selectWeek ?
> 
> @fofi we don't allow selecting a "non-complete" week on the picker, although we could add the verification just in case we change that in the future so everything works the same way, wdyt?

Whatever you think is best, on my side is enough

fofi (approved):

PR #99, opened 2021-10-14
Feature/core 9190 calendar bug

dletran:


dletran:
@Navas23 it could be a bit weird because with the current solution they would depend on the timezone of the machine running the tests

PR #102, opened 2021-12-07
Feature/core 9223 changes multiselect

dletran:


src/elements/GeoSelect/GeoSelect.mixin.js
```diff
@@ -324,18 +334,48 @@ export default {
...
-        end: this.pageSize ? this.lastVisiblePage * this.pageSize : this.filteredOptions.length
+        end,
+        endGroup
       }
     },
 
     visibleOptions () {
-      return this.filteredOptions.slice(this.visibleChunkRange.start, this.visibleChunkRange.end)
+      if (this.hasToCalculatePaginatedGroupItems) {
+        const visibleGroups = _.slice(_.cloneDeep(this.filteredOptions), this.visibleChunkRange.start, this.visibleChunkRange.endGroup)
```
fofi, line 358: Why do we need a clone deep?
Also if we need it, is better if we do it after the slice, so we clone less items

src/elements/GeoSelect/GeoSelect.mixin.js
```diff
@@ -324,18 +334,48 @@ export default {
...
+        const lastGroup = _.last(visibleGroups)
+        lastGroup.items = _.slice(lastGroup.items, this.visibleChunkRange.start, this.visibleChunkRange.end)
```
fofi, line 360: What is exactly what we are slicing here?

fofi: Why not using endGroup here?

fofi (commented):
Please do not forget to add tests when updating code like this!

src/elements/GeoSelect/GeoSelect.mixin.js
dletran: we're slicing the items from the group that should be visible, so here we use the calculated "end" which refers to the last index of the group that is visible

src/elements/GeoSelect/GeoSelect.mixin.js
dletran: we use a clone deep because we're going to slice the items inside a group but we don't want to modify the original object

I'll move the clone after the slice 👌

src/elements/GeoSelect/GeoSelect.mixin.js
```diff
@@ -324,18 +324,48 @@ export default {
...
     visibleChunkRange () {
+      let end = this.pageSize ? this.lastVisiblePage * this.pageSize : this.filteredOptions.length
+      let endGroup
+      if (this.hasToCalculatePaginatedGroupItems) {
+        _.forEach(this.filteredOptions, (group, index) => {
+          const accumulatedItems = _.sumBy(_.slice(this.filteredOptions, 0, index + 1), 'items.length')
+          if (end <= accumulatedItems) {
+            endGroup = index + 1
+            end = end - (accumulatedItems - group.items.length)
+            return false
```
fofi, line 335: I think here we could just return, as the false is not doing anything, is just going to the next item on the loop.

Actually we could remove the return statement also

fofi (approved):

src/elements/GeoSelect/GeoSelect.mixin.js
dletran: the return false is like a break on the forEach to stop looping

fofi: Oh nice, didn't knew that, thanks!

PR #111, opened 2022-09-15
Feature/core 8803 improve geo list item

irmakyz:


src/styles/elements/GeoList/_geo-list-item.scss
```diff
@@ -59,6 +64,13 @@
   align-items: center;
 }
 
+.geo-list-item__icon-and-label__label {
```
dletran, line 67: let's try to order the props alphabetically, I'll create a task to include the lint rules to check that in the DS

dletran (commented):
what about the places where we use geo-list-item without applying these extra styles? have you checked if they also look fine now?

src/styles/elements/GeoList/_geo-list-item.scss
irmakyz: okayy, right i forgot it, fixing

irmakyz:
> what about the places where we use geo-list-item without applying these extra styles? have you checked if they also look fine now?

yess i checked all of them

dletran (approved):

fofi (approved):

PR #113, opened 2022-10-25
Feature/core 3958 tests for chart labels

irmakyz:
I tried to find a way to check the exact positions, however since it is wrapper i think there is no actual values for translations :/ do you have any idea how can i achieve it, i checked other files but in general they are based on values as well

### DIFF
--- a/test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
+++ b/test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
@@ -118,7 +118,7 @@ describe('GeoChartLabels', function () {
     const cssClassFn = (original) => [...original, 'test-bar']
 
     testHorizontalDimension(linearAxisConfig, categoricalAxisConfig, highlightedSegments, cssClassFn, labelGroup)
-    // testVerticalDimension(categoricalAxisConfig, linearAxisConfig, highlightedSegments, null, labelGroup)
+    testVerticalDimension(categoricalAxisConfig, linearAxisConfig, highlightedSegments, cssClassFn, labelGroup)
   })
 
   function testHorizontalDimension (verticalAxis, horizontalAxis, highlightedSegments, cssClassFn, labelGroup) {
@@ -228,10 +228,10 @@ describe('GeoChartLabels', function () {
         data: _.map(highlightedSegments, (value, index) => {
           return {
             labels: [{
-              text: value,
+              text: _.toString(value),
               cornerRadius: 5
             }],
-            [axisConfig.linearAxisConfig.keyForValues]: value,
+            [axisConfig.linearAxisConfig.keyForValues]: _.toString(value),
             [axisConfig.categoricalAxisConfig.keyForValues]: mockDomain[index]
           }
         }),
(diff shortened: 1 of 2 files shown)


dletran:
> I tried to find a way to check the exact positions, however since it is wrapper i think there is no actual values for translations :/ do you have any idea how can i achieve it, i checked other files but in general they are based on values as well

maybe we can do something similar as you're doing to check the text value
1. you get the label element
2. you get the bar element for that label
3. you check that the distance between both is the expected according to the config

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,433 @@
...
+          propsData: {
+            config: barConfigForMultipleLabels
+          }
+        })
+
+        flushD3Transitions()
+        expect(wrapper.find('.geo-chart').exists()).toBe(true)
+        expect(wrapper.find('.geo-chart .geo-chart-bars-group').exists()).toBe(true)
+        _.forEach(mockDomain, (category, index) => {
+          expect(wrapper.find(`.geo-chart-label-group--${index}`).findAll('.geo-chart-labels-group__single-label').length).toEqual(2)
```
dletran, line 206: since we are checking several things in the list of label group we could save the findAll result in a const and then access it in the following lines so we don't repeat the findAll

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,433 @@
...
```
dletran, line 314: same thing here as mentioned before about the findAll

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,433 @@
...
```
dletran, line 423: also here

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,433 @@
...
+  stubGetComputedTextLengthFactory
+} from '../GeoChart/GeoChart.spec-utils' // This has to be imported before D3
+import { createLocalVue, mount } from '@vue/test-utils'
+import GeoChart from '@/elements/GeoChart/GeoChart.vue'
+
+const localVue = createLocalVue()
+localVue.component('geo-chart', GeoChart)
+
+const mockDomain = _.times(3, i => `Bucket ${i}`)
+const highlightedSegments = [10, 15, 20]
```
dletran, line 18: what is this exactly?

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -14,8 +14,8 @@ import GeoChart from '@/elements/GeoChart/GeoChart.vue'
 const localVue = createLocalVue()
 localVue.component('geo-chart', GeoChart)
 
-const mockDomain = _.times(3, i => `Bucket ${i}`)
-const highlightedSegments = [10, 15, 20]
+const categoricalAxisValues = _.times(3, i => `Bucket ${i}`)
+const linerAxisValues = [10, 15, 20]
```
irmakyz, line 18: i changed the naming to be more clear @dletran but in short, they are the values of the axises

fofi:
What are the changes on this pr? Are we adding anything?

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,438 @@
...
+  stubGetComputedTextLengthFactory
+} from '../GeoChart/GeoChart.spec-utils' // This has to be imported before D3
+import { createLocalVue, mount } from '@vue/test-utils'
+import GeoChart from '@/elements/GeoChart/GeoChart.vue'
+
+const localVue = createLocalVue()
+localVue.component('geo-chart', GeoChart)
+
+const categoricalAxisValues = _.times(3, i => `Bucket ${i}`)
+const linerAxisValues = [10, 15, 20]
```
Navas23, line 18: this should be "linearAxisValues", right?

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,438 @@
...
+  })
+
+  afterEach(function () {
+    stubGetBBox.teardown()
+    stubGetComputedTextLength.teardown()
+    stubCreateSVGPoint.teardown()
+    stubGetScreenCTM.teardown()
+  })
+
+  afterEach(function () {
```
Navas23, line 108: Why do we need this second afterEach?

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,438 @@
...
+        expect(wrapper.find('.geo-chart').exists()).toBe(true)
+        expect(wrapper.find('.geo-chart .geo-chart-bars-group').exists()).toBe(true)
+        _.forEach(categoricalAxisValues, (category, index) => {
+          const singleLabelGroups = wrapper.find(`.geo-chart-label-group--${index}`).findAll('.geo-chart-labels-group__single-label')
+
+          expect(singleLabelGroups.length).toEqual(2)
+          expect(singleLabelGroups.at(0).text()).toEqual('<<')
+          expect(singleLabelGroups.at(1).text()).toEqual(_.toString(category))
+        })
+        expect(wrapper.find('.geo-chart-labels-group-container').exists()).toBe(true)
```
Navas23, line 212: I think we should check this before the labels

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,438 @@
...
+        }
+        const barConfigForMultipleLabels = _.assign({}, barConfig, { labelGroups: [multipleLabelsGroup] })
+        const wrapper = mount(GeoChart, {
+          propsData: {
+            config: barConfigForMultipleLabels
+          }
+        })
+
+        flushD3Transitions()
+        expect(wrapper.find('.geo-chart').exists()).toBe(true)
```
Navas23, line 312: Why we are checking this only in some tests?

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,438 @@
...
+        expect(wrapper.find('.geo-chart .geo-chart-bars-group').exists()).toBe(true)
+        _.forEach(categoricalAxisValues, (category, index) => {
+          const singleLabelGroups = wrapper.find(`.geo-chart-label-group--${index}`).findAll('.geo-chart-labels-group__single-label')
+
+          expect(singleLabelGroups.length).toEqual(2)
+          expect(singleLabelGroups.at(0).text()).toEqual('<<')
+          expect(singleLabelGroups.at(1).text()).toEqual(_.toString(category))
+        })
+        expect(wrapper.find('.geo-chart-labels-group-container').exists()).toBe(true)
```
Navas23, line 321: Same here about doing this before checking the labels

irmakyz:
i added new tests and looking at comments rn, will let you know when it is ready to review again @Navas23 @dletran

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
irmakyz: okii

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
irmakyz: it is not needed anymore, i added it before to trying to mock some functions, deleting now

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
irmakyz: hmmm yess makes sense

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
irmakyz: i deleted it now, it was unnecessary

irmakyz:
it is ready to review again @Navas23 @dletran

src/elements/GeoChart/GeoChartConfigs/GeoChartConfigAdapter.mixin.js
```diff
@@ -101,7 +101,6 @@ export default {
           const defaultPadding = isHorizontal
             ? { top: 0, bottom: 0, left: 15, right: 10 }
             : { top: 0, bottom: 0, left: 0, right: 0 }
-
```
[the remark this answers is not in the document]
irmakyz: ah oki reverting them

Navas23 (approved):

Navas23:
Don't forget to update the readme and the version

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
```diff
@@ -0,0 +1,499 @@
...
+
+        flushD3Transitions()
+        // since there is no comparison, should be wide bar charts
+        expect(wrapper.find('.geo-chart-labels-group__single-label').attributes().class).toContain('geo-chart-value-label--medium')
+        _.forEach(linearAxisValues, (value, index) => {
+          expect(wrapper.find(`.geo-chart-label-group--${index} text`).text()).toEqual(_.toString(value))
+        })
+        wrapper.destroy()
+      })
+      it('Should render the labels with correct positionings', () => {
```
dletran, line 270: we're not actually checking the positioning here right? we can use the same name as for vertical dimensions test: 'Should render the labels with correct classes'

dletran (commented):
Just one small comment

test/unit/specs/elements/GeoChartLabel/GeoChartLabel.spec.js
irmakyz: yess true true, fixing right now

dletran (approved):

PR #115, opened 2022-11-21
Feature/core 10256 improve geo tree

Navas23:


src/elements/GeoTree/GeoTree.vue
```diff
@@ -145,9 +145,9 @@ export default {
       default: () => ({})
     },
     /**
-     * Initial categories to be expanded, with truthy/falsy category ids
+     * Initial categories to be expanded, with truthy/falsy category ids, it's being watched to react to outside changes
      */
-    initialExpandedCategories: {
+    dynamicExpandedCategories: {
```
fofi, line 150: If we replace a property, then we should change the mayor version as is not retrocompatible

fofi: Maybe we can keep both, wdyt?

dletran (commented):
LGTM just Fofi's comment

src/elements/GeoTree/GeoTree.vue
Navas23: Sure, I updated as minor instead of a patch but a major is the right thing here

About keeping both... which benefit do you see? I think with 1 cover the use cases we can need

fofi: We could keep it if we want to be retrocompatible or if we want to support the same case but without a watch.

Maybe is not worth it, I was just asking

Navas23: For me is not worth it if we are not going to use it, we can add it easily if we need it later

fofi: Oki, lets just change the version to major

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+
+## 30.1.0
```
fofi, line 4: Change version here and add the property we add and remove if possible

fofi (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+New:
```
dletran, line 5: should we add 'Breaking' also to specify the breaking changes (even though is the only change)?

dletran (approved):

CHANGELOG.md
Navas23: Yeah it does not harm

PR #116, opened 2022-12-13
Feature/core 10304 add geo tree customization

Navas23:
Finally I had to use folder prop instead of slots as we would have to also modify geo-list-item to allow icon as slots

### DIFF
--- a/src/elements/GeoTree/GeoTreeItem.vue
+++ b/src/elements/GeoTree/GeoTreeItem.vue
@@ -152,8 +152,8 @@ export default {
       return this.hasChildren
         ? this.useFolderIcon
           ? this.isExpanded
-            ? ['fa', 'folder-open']
-            : ['fa', 'folder']
+            ? ['fal', 'folder-open']
+            : ['fal', 'folder']
           : ['fal', 'chevron-right']
         : null
     },
--- a/src/elements/GeoTree/GeoTreeItem.vue
+++ b/src/elements/GeoTree/GeoTreeItem.vue
@@ -1,8 +1,6 @@
 <template>
   <li
     class="geo-tree-item"
-    @mouseover="isShowActionButton = true"
-    @mouseout="isShowActionButton = false"
   >
     <geo-list-item
       :class="{
@@ -36,8 +34,8 @@
           </geo-tooltip>
         </span>
       </label>
-      <div slot="trailingAccessoryItem">
-        <span v-show="isShowActionButton">
+      <template slot="trailingAccessoryItem">
+        <span>
           <slot name="trailingAccessoryAction" />
         </span>
         <input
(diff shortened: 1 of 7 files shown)


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -1,13 +1,12 @@
 <template>
   <li
     class="geo-tree-item"
-    @mouseover="isShowActionButton = true"
```
fofi, line 4: Why do we want to remove this?

Navas23:
The problem of the 2 props is that we could not have those icons in the DS... So I will keep it like this but changing the prop name

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -128,6 +138,14 @@ export default {
     expandedCategories: {
       type: Object,
       required: true
+    },
+    /**
+     * Use folder icon instead of chevron
+     */
+    useFolderIcon: {
```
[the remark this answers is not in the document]
Navas23: WDYT about isUsingFolderIcon

src/elements/GeoTree/GeoTreeItem.vue
fofi: And why before we had this? are we going to break something?

Navas23: IDK, we were not using it (in fact it was not usable as we had no props in the slots so we could not know which element we clicked)

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -128,6 +141,26 @@ export default {
...
+    },
+    /**
+     * Optional Font Awesome 5 icon to use as collapsed icon
+     *
+     * See [vue-fontawesome](https://www.npmjs.com/package/@fortawesome/vue-fontawesome#explicit-prefix-note-the-vue-bind-shorthand-because-this-uses-an-array)
+     * for more info about this.
+     */
+    collapsedIcon: {
+      type: Array,
+      required: false
```
fofi, line 153: We could add here the defaultIcon

fofi (approved):

src/elements/GeoTree/GeoTreeItem.vue
Navas23: I didn't add it so I can know if the icon is custom or not (to rotate the icon)

fofi: But, we will rotate the icon if we do not have expandedIcon, right? it has nothing todo with the collapsed one

Navas23: You are right, I will update it

dletran (approved):

fofi (approved):

PR #117, opened 2022-12-22
Feature/core 8922 dropdown issue end of screen

irmakyz:


src/elements/GeoDropdown/GeoDropdown.vue
```diff
@@ -335,6 +335,10 @@ export default {
       const fitsBelow = popupRect.height < availableHeightBelow
       const fitsAbove = popupRect.height < availableHeightAbove
 
+      if (fitsAbove && (containerRect.top + availableHeightAbove > viewport.height)) {
```
dletran, line 338: what does this condition mean? let's try to include the second part `(containerRect.top + availableHeightAbove > viewport.height)` in a constant so it's easier to read

dletran: should we make this check for the 4 sides? top, bottom, left and right?

irmakyz: yes i think i need to add for bottom side as well. this is a condition that if the popup is higher then the container and excessing the end of screen, it detects and making the popup smaller and handles the translation

irmakyz: - [ include second part in the constant]

PR #118, opened 2022-12-23
Feature/core 10308 allow empty folders geo tree

Navas23:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -210,19 +213,22 @@ export default {
     isExpanded () {
       return !!this.expandedCategories[this.category[this.keyForId]]
     },
-    hasChildren () {
-      return this.totalCategoryChildren > 0
+    isSingle () {
```
dletran, line 216: what do you think about renaming this `isChild` to keep the children/parent naming or `isSingleItem` to clarify more?

dletran (approved):

src/elements/GeoTree/GeoTreeItem.vue
Navas23: I prefer isSingleItem, because a folder and an item can be siblings, not only parent/children, wdyt?

dletran: sounds good!

fofi (approved):

PR #119, opened 2023-01-16
Feature/core 10436 fix empty folders geo tree

Navas23:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -195,8 +195,9 @@ export default {
       if (this.isInputDisabled) return false
 
       const allAreChildrenSelected = category => {
-        return _.every(category[this.keyForSubcategory], subCategory => {
-          if (subCategory[this.keyForSubcategory]) return allAreChildrenSelected(subCategory)
+        const selectableChildren = this.getSelectableChildrenForCategory(category)
+        return _.every(selectableChildren, subCategory => {
+          if (_.size(subCategory[this.keyForSubcategory])) return allAreChildrenSelected(subCategory)
```
fofi, line 200: I think we don't need to check the size, as it if has the prop, then it has a size

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -210,6 +211,14 @@ export default {
 
       return sumOfChildren(this.category)
     },
+    totalSelectableCategoryChildren () {
+      const sumOfSelectableChildren = category => {
+        const selectableChildren = this.getSelectableChildrenForCategory(category)
```
fofi, line 216: We could calculate the selectable children tree once, but probably is not worth it

fofi (approved):

src/elements/GeoTree/GeoTreeItem.vue
Navas23: Only cause we are checking it in line 198, in other case it could be an empty array

src/elements/GeoTree/GeoTreeItem.vue
Navas23: IMO is not, I though about it and decided not to

dletran (approved):
LGTM

PR #121, opened 2023-02-07
Feature/core 10456 geo tree improvements

dletran:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -11,10 +11,12 @@
       @click="handleClick"
     >
       <label class="geo-tree-item__label">
-        <geo-highlighted-string
-          :highlighted-chars="category.matches"
-          :reference-string="category[keyForLabel]"
-        />
+        <geo-trimmed-content>
```
Navas23, line 14: This can have an impact in performance, have you checked in the app?

Navas23 (commented):
Looks good just one comment

src/elements/GeoTree/GeoTreeItem.vue
dletran: yep I had this concern too, but if we don't include this we won't be able to see the full name of items with long names

That said, I checked with several items and it looks good, I think it will only be a problem if there are many many items displaying at once (like in a table with many cells) that have a long name which in this case shouldn't happen too often

Navas23 (approved):

fofi (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 32.0.4
+Fix:
+- `GeoTree` and `GeoTreeItem`: Fix description in subcategories, allow custom sorting and added trimmed content to items labels
```
Navas23, line 5: Change this trimmed to marquee pls

src/styles/elements/GeoTree/_geo-tree-item.scss
```diff
@@ -7,11 +7,9 @@
   cursor: pointer;
   display: flex;
   gap: 2px;
-  .highlighted-string__container {
-    flex-shrink: 1;
+  .geo-trimmed-content {
```
Navas23, line 10: We don't need this anymore, right?

dletran: true! I'll remove it

src/styles/elements/GeoTree/_geo-tree-item.scss
```diff
@@ -3,15 +3,8 @@
 }
 
 .geo-tree-item__label {
-  align-items: center;
```
fofi, line 6: Its fine to remove all these?

CHANGELOG.md
fofi: Update this

src/styles/elements/GeoTree/_geo-tree-item.scss
dletran: we need to remove them to have the same "styles logic" as in general markers, otherwise the number of items inside the category is going to appear aligned to the right of the label and doesn't look so good

fofi (approved):

PR #122, opened 2023-02-14
Feature/core 10473 folder events geo tree

Navas23:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -244,19 +245,25 @@ export default {
...
+      this.$emit('check-folder', category, isChecked, isDelegated)
     },
     /**
      * To check all items of a category
      */
-    handleCheck (category, isChecked) {
+    handleCheck (category, isChecked, isDelegated = false) {
       if (category[this.keyForSubcategory]) {
+        if (_.size(category[this.keyForSubcategory])) { // We don't want to emit event for empty folders
+          this.$emit('check-folder', _.omit(category, this.keyForSubcategory), isChecked, isDelegated)
```
fofi, line 260: Why do we want to omit the keyForSubcategory ?

fofi (approved):

src/elements/GeoTree/GeoTreeItem.vue
Navas23: We could we sending a huge object there that we are not going to use, IMO it doesn't make sense to send all the subfolder and subitems

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 33.0.0
+New:
```
dletran, line 4: small thing: this is a breaking change not just new right?

dletran (approved):

CHANGELOG.md
Navas23: It's breaking and new, I will update it

src/elements/GeoTree/GeoTreeItem.vue
irmakyz: makes sense for me

PR #123, opened 2023-03-03
Add support for nodejs 18

SteveFlames:


package.json
```diff
@@ -178,7 +178,7 @@
     "webpack-merge-and-include-globally": "^2.1.16"
   },
   "engines": {
-    "node": "^14.0.0",
+    "node": ">= 12.0.0",
```
Navas23, line 181: Do we support building with node12?

fofi (approved):

package.json
SteveFlames: Yep, we are even testing with node12.
https://github.com/geoblink/design-system/pull/123/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R24

.travis.yml
```diff
@@ -28,6 +28,10 @@ jobs:
       node_js: 14
       script:
         - yarn test
+    - stage: test
```
SteveFlames, line 31: We could run tests with nodejs16 also

dletran (approved):

Navas23 (approved):

PR #125, opened 2023-03-14
Feature/core 10313 tree drag

Navas23:


src/elements/GeoTree/GeoTreeMixin.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+export default {
+  props: {
+    /**
+     * Group object in case you want to support drag elements
+     *
+     * See [vuedraggable](https://www.npmjs.com/package/vuedraggable)
+     * for more info about this.
+     */
+    group: {
```
fofi, line 12: I think this name is not clear enough that is ment to be for the draggable functionality
maybe `draggableGroup` or something like that?

src/elements/GeoTree/GeoTreeMixin.js
```diff
@@ -0,0 +1,44 @@
...
+      return !this.group.pull
+        ? 'geo-tree__drag-disabled'
+        : ''
+    }
+  },
+  methods: {
+    startDrag ($event) {
+      this.$emit('start-drag', $event)
+    },
+    endDrag ($event) {
```
fofi, line 31: Add blank lines between methods

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -56,45 +56,70 @@
...
-        :expanded-categories="expandedCategories"
-        :checked-items="checkedItems"
-        :collapsed-icon="collapsedIcon"
-        :expanded-icon="expandedIcon"
-        :description-icon="descriptionIcon"
-        @check-item="handleCheckChildItem"
-        @check-folder="handleCheckChildFolder"
-        @click="handleClick"
-        @toggleExpand="toggleExpand"
+      <draggable
```
fofi, line 59: Is there any change in terms of performance due this change?
I mean, because maybe now we are adding more stuff to the DOM even if we don't need the draggable thing

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -56,45 +56,70 @@
...
+          :category="subcategory"
+          :key-for-id="keyForId"
+          :key-for-label="keyForLabel"
+          :key-for-subcategory="keyForSubcategory"
+          :expanded-categories="expandedCategories"
+          :checked-items="checkedItems"
+          :collapsed-icon="collapsedIcon"
+          :expanded-icon="expandedIcon"
+          :description-icon="descriptionIcon"
+          :group="group"
```
fofi, line 85: Why does the geo-tree-item needs the group prop?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -56,45 +56,70 @@
...
+        @change="changeDrag($event, category[keyForSubcategory])"
       >
-        <template
-          slot="trailingAccessoryAction"
-          slot-scope="{ item }"
+        <geo-tree-item
+          v-for="subcategory in category[keyForSubcategory]"
+          :key="subcategory[keyForId]"
+          :class="dragClassToIgnore"
+          :data-test="`subcategory-${subcategory[keyForId]}`"
```
fofi, line 75: Why do we have a test property?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -56,45 +56,70 @@
...
-        @click="handleClick"
-        @toggleExpand="toggleExpand"
+      <draggable
+        :list="category[keyForSubcategory]"
+        :group="group"
+        :sort="false"
+        :disabled="!group"
+        :filter="`.${dragClassToIgnore}`"
+        drag-class="geo-tree__dragger"
+        ghost-class="geo-tree__ghost"
```
fofi, line 66: Should we make it more clear that these classes are for draggable?

src/elements/GeoTree/GeoTreeItem.vue
Navas23: It was already here, maybe this was not the best way to test this component

src/elements/GeoTree/GeoTreeItem.vue
Navas23: Right now they are for internal use, but we can rename them if you think it's better

src/elements/GeoTree/GeoTreeItem.vue
Navas23: I haven't seen any, we are just adding a wrapper div

src/elements/GeoTree/GeoTreeItem.vue
fofi: Are we going to use them outside the DS?

src/elements/GeoTree/GeoTreeItem.vue
fofi: Check if we can remove it, or if we need it please, if we need it, we could create a task to change it a bit if we need

src/elements/GeoTree/GeoTreeMixin.js
fofi: But when using the geo-tree, is a bit weird in my opinion, is more clear if we add "draggable" to the name

Navas23: Ok, I will rename it

src/elements/GeoTree/GeoTreeItem.vue
Navas23: We need it just for the tests. I'm gonna create a task to improve it

src/elements/GeoTree/GeoTreeItem.vue
Navas23: I'm using them as a ñapa, they are not really "needed"

fofi: Which ñapa?

dletran (approved):

fofi (approved):

PR #127, opened 2023-03-30
Add support for single-select-mode

Navas23:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -42,11 +42,12 @@
           />
         </span>
         <input
+          v-if="!isSingleSelectMode || isSingleItem"
```
dletran, line 45: this is to not show the input on folders when it's single select mode right?

dletran (approved):

src/elements/GeoTree/GeoTreeItem.vue
Navas23: Exactly

CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
 # Changelog
 
+## 33.2.0
+New:
+- `GeoTree` and `GeoTreeItem`: Add support for single select mode UX. Added prop `is-single-select-mode` to use it
```
fofi, line 5: LEts update this comment and prop comment to make more clear that the new prop is to use radio vs checkbox input

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -42,11 +42,12 @@
...
           :id="category[keyForId]"
           :checked="isChecked"
           :indeterminate.prop="isIndeterminate"
```
fofi, line 48: Not sure if a good idea, but we could stop calculating some computed properties that do not make sense on single mode

Navas23: IMO is not worth it for this case

fofi (approved):

PR #128, opened 2023-04-25
Revert changes for node18

stefanosgk:


dletran (commented):
should we update also the changelog for this version?

SteveFlames:
> should we update also the changelog for this version?

hmm, to mention that it was reverted so that we don't get confused in >34.0.0?

dletran:
> > should we update also the changelog for this version?
> 
> hmm, to mention that it was reverted so that we don't get confused in >34.0.0?

yeah I mean if we're reverting the changes without releasing a new version then the changelog is not real anymore so we should clarify that IMO

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 18
+  - 14
```
fofi, line 3: Why not using 16 also here?

fofi (changes requested):
Lets also change the version and changelog

dletran (approved):

PR #132, opened 2023-08-10
Feature/core 10913 fix radio button geo tree item

irmakyz:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -42,7 +42,15 @@
           />
         </span>
         <input
-          v-if="!isItemInputHidden"
+          v-if="isSingleSelectMode && isSingleItem"
```
Navas23, line 45: Why do we need to have 2 inputs?

src/elements/GeoTree/GeoTree.examples.md
```diff
@@ -125,6 +125,137 @@ export default {
 </script>
 ```
 
+### Basic example with single select
+
+```vue live
+<template>
```
Navas23, line 131: Why do we need another single select example and not use the current one?

Navas23 (commented):
I don't understand the solution, why not stopping the propagation is the solution?

src/elements/GeoTree/GeoTreeItem.vue
irmakyz: hey, because i couldn't find a way to set @click.stop as a conditional way,  so did as this way

src/elements/GeoTree/GeoTree.examples.md
irmakyz: all examples are based on checkboxes, so i added one with radio input

irmakyz:
> I don't understand the solution, why not stopping the propagation is the solution?

i searched on the internet and it says when we stop the event propagation, it effects the functionality of the radio inputs, that's why i removed it and it works now

src/elements/GeoTree/GeoTree.examples.md
Navas23: There is one with radio inputs

src/elements/GeoTree/GeoTreeItem.vue
Navas23: But what is happening if we keep the click.stop? the event is not being sent? have you checked the input event instead of the click one?

src/elements/GeoTree/GeoTree.examples.md
irmakyz: my bad deleting this one

src/elements/GeoTree/GeoTree.examples.md
```diff
@@ -800,134 +800,4 @@ export default {
   }
 }
 </script>
-```
-
-### With max number of items checked
```
Navas23, line 805: Why did you remove this one?

src/elements/GeoTree/GeoTreeItem.vue
irmakyz: yes i checked both of them, the event is not being sent

Navas23: And why don't we capture the click normally and if it's singleSelectMode we stop the propagation?

irmakyz: hmm, yeah is a better idea maybe, i will change

irmakyz: - [ ] capture the click normally and if it's not singleSelectMode we stop the propagation

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -42,13 +42,13 @@
           />
         </span>
         <input
-          v-if="!isItemInputHidden"
+          v-if="!isInputDisabled"
```
Navas23, line 45: Why this change? this will break stuff for sure

irmakyz: - [ ] fix wrong change

Navas23 (approved):

dletran (approved):

PR #139, opened 2023-12-05
add second level load more button

irmakyz:


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -302,6 +325,18 @@ export default {
       if (this.isSingleItem) return false
 
       return this.isSingleSelectMode || this.isFolderSelectHidden
+    },
+
+    hasMoreResultsToLoad () {
+      if (!this.hasLoadMoreButton) return false
+
+      return this.pageSize * this.visiblePages < this.category[this.keyForSubcategory].length
```
Navas23, line 333: Let's use lodash to get the size

Navas23 (changes requested):
With your solution we are only applying it to 1 sublevel

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@geoblink/design-system",
-  "version": "35.5.0-beta.3",
+  "version": "35.5.0-beta.5",
```
Navas23, line 3: Why do we need 5 betas to test this? should be straightforward to test it in local

Navas23 (commented):
Don't we need to pass the slot to the following geo-tree-item?

package.json
irmakyz: first 2 betas were used before, so i started from 2, i fixed the sub level issue and then saw your comment so i published another one, that's why it jumped from 3 to 5

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -105,6 +107,12 @@
             />
           </template>
         </geo-tree-item>
+        <geo-list-footer-button
+          v-if="hasMoreResultsToLoad"
+          @click="loadNextPage"
+        >
+          <slot name="moreItemResultsTextContent" />
```
dletran, line 114: I feel like we're mixing names in several places, I think we should either stick with `results` or `items` so we either have `moreResultsTextContent` or `moreItemsTextContent`

So if the flag is going to be `hasMoreResultsToLoad`, I'd say to use results also, wdyt?

Navas23:
> With your solution we are only applying it to 1 sublevel

Have you checked this? is this what we want?

irmakyz:
> > With your solution we are only applying it to 1 sublevel
> 
> Have you checked this? is this what we want?

yes, i fixed it you can see in the commit

irmakyz:
> Don't we need to pass the slot to the following geo-tree-item?

yes, i passed it after your comments on tuesday

src/elements/GeoTree/GeoTreeItem.vue
dletran: as I mentioned I'd only use results or items but not both, so if we want to tell them apart let's use `moreItemsTextContent`

irmakyz: okay then i will use moreItemsTextContent and moreCategoriesTextContent
is it better?

irmakyz: - [ ] rename text contents

dletran (approved):

fofi (approved):

Navas23 (approved):

PR #142, opened 2023-12-28
rename classnames according to the bem notation

irmakyz:


Navas23 (commented):
This new classes are not following BEM notation. Check this and ask me if it's not clear https://getbem.com/naming/

Also this should not be a major change, IMO it should be a minor one

irmakyz:
> This new classes are not following BEM notation. Check this and ask me if it's not clear https://getbem.com/naming/
> 
> Also this should not be a major change, IMO it should be a minor one

- [ ] fix classes according to BEM notation
- [ ] Change it as a minor change

src/elements/GeoList/GeoListGroup.vue
```diff
@@ -5,26 +5,26 @@
       class="geo-list-group__header"
       @click="emitClick($event)"
     >
-      <div class="geo-list-group__header__icon-and-label">
+      <div class="geo-list-group__header-icon-and-label">
         <div
           v-if="icon"
-          class="geo-list-group__header__icon-and-label__icon-container"
+          class="geo-list-group__header-icon-and-label__icon-container"
```
Navas23, line 11: With these classes we still have the same problem, right?

irmakyz: but i think it makes sense , icon container is an element

Navas23: Let's discuss everything with the rest when they come back

irmakyz:
i made some changes, i think now it is better, wdyt? @Navas23

Navas23:
Yeah it's better, but let's wait for the rest in order to decide

fofi (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 37.0.0
```
dletran, line 3: Remember to update this

fofi (approved):

PR #147, opened 2025-03-20
Cnl/web 1403 geo tree improvements

Navas23:
Breaking change:
- `GeoTree`: The `isSingleSelectMode` prop is replaced by `itemInputMode` prop.

New:
- `GeoTreeItem`: Add support for custom icons for single items adding the `itemIcon` property to the category object
- `GeoTreeItem`: Add support for disabling single items adding the `disabledInfo` property to the category object


src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -125,17 +134,42 @@
...
   mixins: [GeoTreeMixin],
   props: {
+    /**
+     * Category to be displayed
+     * Structure:
+     * {
+     *     id: String,
+     *     label: String,
+     *     itemIcon?: Array,
+     *     disabledInfo?: String,
```
fofi, line 161: LEts use `disabledLabel` to keep some consistency with the names

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -125,17 +134,42 @@
...
   },
   mixins: [GeoTreeMixin],
   props: {
+    /**
+     * Category to be displayed
+     * Structure:
+     * {
+     *     id: String,
+     *     label: String,
+     *     itemIcon?: Array,
```
fofi, line 160: Why not just `icon`?

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -222,13 +256,19 @@ export default {
...
-      default: false
-    },
+    /**
+     * Mode of the input (take into account the logic still remains on how to handle checked-items prop)
+     *
+     * Supported `itemInputMode` values are exported under `INPUT_MODES` named export.
+     * See [Component Constants](/docs/components-constants.html) for more info on how
+     * to use those constants in your code.
+     */
+    itemInputMode: enumPropertyFactory({
```
fofi, line 266: I think this name is not really explanatory of what it means.

maybe `selectMode` or `selectType`

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -318,18 +358,27 @@ export default {
...
+          return 'checkbox'
+        default:
+          return null
+      }
     },
 
     isItemInputHidden () {
-      if (this.isSingleItem) return false
+      if (this.isSingleItem) {
+        return this.itemInputMode === INPUT_MODES.HIDDEN
```
fofi, line 378: Why do we need hidden as a select type?

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
fofi, line 1: Could we incldue some more extensive tests?

src/elements/GeoTree/GeoTreeItem.vue
Navas23: I will rename it with disabledTooltipText, as it's more verbose

src/elements/GeoTree/GeoTreeItem.vue
Navas23: It only apply to items (elements without children), that's why I specify. But maybe it make sense to allow this also for folders and rename it to customIcon, wdyt?

src/elements/GeoTree/GeoTreeItem.vue
Navas23: For some cases we just want to detect the click and do not display input on elements

src/elements/GeoTree/GeoTreeItem.vue
dletran: if it's straightforward I'd say to do that and use just `icon` for the prop name

src/elements/GeoTree/GeoTreeItem.vue
dletran: I also think `selectMode` is more straightforward, the only thing is that for the "hidden" value it wouldn't make sense
maybe `selectMode = 'single-auto' | 'single' | 'multiple'` ?

package.json
```diff
@@ -183,7 +183,7 @@
     "loader-utils": "^1.4.1"
   },
   "engines": {
-    "node": ">=12.0.0 <17.0.0",
+    "node": ">=20.0.0 <21.0.0",
```
stefanosgk, line 186: why restrict it so much? It should still work for versions lesser than 20, it's just vuepress that's failing to build with those versions.

Navas23: That's why, vuepress can not build and you can not dev on in with other versions, so I think it makes sense to restrict it

src/elements/GeoTree/GeoTreeItem.vue
dletran: ok I get your point, inputType sounds good to me 👍

test/unit/specs/elements/GeoTree/GeoTreeItem.spec.js
fofi: If you think with this is enough, lets keep it as it is

PR #150, opened 2025-08-21
Add rebranding tokens

dletran:
Added rebranding tokens:
- colors
- fonts

Updated the following components using them:
- GeoButton
- GeoPill

Added deprecated flag to old tokens that we want to discontinue once we stop using them in the app, also updated the components we use for the docs to reflect this deprecation status

Further components may need modification but we can update them step by step along with the app rebranding

### DIFF
--- a/src/elements/GeoButton/GeoButton.mixin.js
+++ b/src/elements/GeoButton/GeoButton.mixin.js
@@ -1,4 +1,4 @@
-import { VARIANTS as GeoAlertVariants } from '../GeoActivityIndicator/GeoActivityIndicator'
+import { VARIANTS as GeoActivityIndicatorVariants } from '../GeoActivityIndicator/GeoActivityIndicator'
 
 const TYPES = {
   primary: 'primary',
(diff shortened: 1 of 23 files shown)


src/elements/GeoButton/GeoButton.mixin.js
```diff
@@ -46,15 +45,13 @@ export default {
     activityIndicatorVariant () {
       switch (this.type) {
         case TYPES.primary:
-          return GeoAlertVariants.primary
+          return GeoAlertVariants['dark-transparent']
```
fofi: Was this on purpose?

Navas23 (commented):
Looks good! Just Fofi's comment

src/elements/GeoButton/GeoButton.mixin.js
dletran: it is yes, with the new primary color for the geobutton the variant for the activity indicator was using a color that wasn't really visible

ideally we should revisit all the geoactivity variants though but as I mentioned we can update components step by step while updating the app

fofi: In the docu for GeoAlertVariants, dark-transparent did not appear, I guess we have it, but why is not on the docu?

dletran: I think it's a naming issue, we have:
`import { VARIANTS as GeoAlertVariants } from '../GeoActivityIndicator/GeoActivityIndicator'`
so it should be GeoActivityIndicatorVariants where there is a `dark-transparent` one

I'll change it

fofi (approved):

PR #153, opened 2025-09-15
Vue 3 migration: prerequisites

dletran:
Vue 3 migration:
This is the 1st of 3 PRs:

1. **Migration prerequisites**
2. Lib update
3. Doc tool migration

In this one we're applying the pre-requisites before updating Vue to Vue 3, those are:
- Update lo latest Vue 2 version
- Migrate functional components to standard ones
- Migrate slots syntax to new syntax
- Migrate $listeners and $props to new $attrs


src/elements/GeoAlert/GeoAlert.vue
```diff
@@ -1,16 +1,15 @@
-<template functional>
+<template>
   <div
-    v-bind="data.attrs"
+    v-bind="$attrs"
     :class="[
-      data.class,
-      data.staticClass,
+      $attrs.class,
```
fofi: Why is needed to send classes down?

package.json
```diff
@@ -12,11 +12,11 @@
...
     "start": "run-s docs:build:markdown docs:watch",
     "dev": "run-s start",
-    "docs": "vuepress dev --open",
-    "docs:dev": "vuepress dev",
+    "docs": "NODE_OPTIONS='--openssl-legacy-provider' vuepress dev --open",
+    "docs:dev": "NODE_OPTIONS='--openssl-legacy-provider' vuepress dev",
     "docs:watch": "run-p -l theo:watch docs:watch:markdown docs",
     "docs:build": "run-s theo:build:docs docs:build:markdown docs:build:vuepress",
-    "docs:build:vuepress": "vuepress build",
+    "docs:build:vuepress": "NODE_OPTIONS='--openssl-legacy-provider' vuepress build",
```
fofi, line 19: Why the change?

src/elements/GeoCalendar/GeoCalendar.vue
```diff
@@ -30,7 +30,6 @@
           <!-- @slot Use this slot to customize the message shown when there is an error in one of the selected dates -->
           <geo-input-message
             v-if="showFromFormatError"
-            slot="message"
```
fofi, line 33: Why removing this sloT?

src/elements/GeoAlert/GeoAlert.vue
dletran: tbh I wasn't completely sure, I think we can probably just remove it but there were so many thing to check that I went with the quick fix

src/elements/GeoTree/GeoTreeItem.vue
```diff
@@ -395,6 +392,7 @@ export default {
     }
   },
   beforeMount () {
+    this.tooltipAlignment = ALIGNMENTS.start
```
fofi, line 395: Are we using this?

package.json
dletran: I was having problems with my node and I had to add this, anyway I committed it by mistake but I remove it in the 3rd PR when I replace it with vitepress

src/elements/GeoDropdown/GeoDropdown.vue
```diff
@@ -8,7 +8,12 @@
     }"
   >
     <!-- @slot Use this slot to customize the button used to toggle the popup -->
-    <slot name="toggleButton" />
+    <div
+      ref="toggleButtonWrapper"
+      class="geo-dropdown__toggle-button-wrapper"
```
fofi, line 20: Why this change?

src/elements/GeoCalendar/GeoCalendar.vue
dletran: Because this slot is not really applying to anything, this component is inside a `<div>` which can't have slots, I think we probably changed this template and we kept this accidentally

src/elements/GeoCalendar/GeoCalendarDropdown.vue
```diff
@@ -6,78 +6,82 @@
...
-          :previous-date-in-selected-granularity-icon="previousDateInSelectedGranularityIcon"
-          :to-input-placeholder="toInputPlaceholder"
-          :is-from-date-disabled="isFromDateDisabled"
-          :is-to-date-disabled="isToDateDisabled"
-          @emit-from-date="emitFromDate"
-          @emit-to-date="emitToDate"
+    <template #toggleButton>
+      <slot
+        name="toggleButton"
+        :toggleCalendarPopup="toggleCalendarPopup"
```
fofi: Is this change expected?

fofi (commented):
Just some small comments! looks good!

src/elements/GeoTree/GeoTreeItem.vue
dletran: yep, we were using it in the template before but not defining it correctly here

src/elements/GeoDropdown/GeoDropdown.vue
dletran: this is a tricky one: this.$slots doesn't work the same way in vue 2.7 and vue 3, so while using the vue-compat library the method we had to calculate this.toggleButtonWidth wasn't working correctly (sometimes it didn't return the content inside the slot)

To have a more consistent behavior I added a div that has `display: contents` as part of its styling which causes an element's children to appear as if they were direct children of the element's parent, i.e: it's a div wrapper that acts as if there was no wrapper. The thing is that this wrapper allows us to put a ref on it and directly calculate its width

src/elements/GeoCalendar/GeoCalendarDropdown.vue
dletran: yes, due to vue components not being eligible to have the `slot` prop anymore, we have to change the HTML structure to fit a template, in this case the restructuring was a bit more complex than other cases to keep the same behaviour

src/elements/GeoDropdown/GeoDropdown.vue
fofi: Can we include a comment on the code so we do not forget about this?

src/elements/GeoCalendar/GeoCalendarDropdown.vue
fofi: Why the prop does not has snake-case anymore?

src/elements/GeoDropdown/GeoDropdownSelectButton.examples.md
```diff
@@ -188,4 +192,4 @@ export default {
   }
 }
 </script>
-```
\ No newline at end of file
+````
```
Navas23: Is this on purpose?

src/elements/GeoSegmentedControl/GeoDangerSegmentedControlItem.vue
```diff
@@ -1,16 +1,12 @@
-<template functional>
+<template>
   <geo-segmented-control-item
-    :ref="data.ref"
-    v-bind="data.attrs"
-    :class="[
-      data.class,
-      data.staticClass
```
Navas23, line 7: It's fine to lose staticClass here?

src/elements/GeoSegmentedControl/GeoInfoSegmentedControlItem.vue
```diff
@@ -1,16 +1,12 @@
-<template functional>
+<template>
   <geo-segmented-control-item
-    :ref="data.ref"
-    v-bind="data.attrs"
-    :class="[
-      data.class,
-      data.staticClass
```
Navas23, line 7: Same here about static class

Navas23 (commented):
Just small comments appart from Fofi's pending ones

src/elements/GeoCalendar/GeoCalendarDropdown.vue
dletran: I'll check that

src/elements/GeoDropdown/GeoDropdownSelectButton.examples.md
dletran: nope I'll remove it, eagle eye 👀

src/elements/GeoSegmentedControl/GeoDangerSegmentedControlItem.vue
dletran: I'll recheck and confirm, I think it's fine to drop it everywhere, also in the other place fofi mentioned because when applying a class to a component directly it gets applied without having to pass it down

src/elements/GeoAlert/GeoAlert.vue
dletran: we can definitely remove it because:
1. when binding $attrs we already include this
2. vue 3 already apply custom classes to the parent element of the component so no need to bind it separately

src/elements/GeoSegmentedControl/GeoDangerSegmentedControlItem.vue
dletran: it's ok to remove it, as I mentioned in a previous comment this would be redundant because they're included in $attrs and we're already binding them, also it's not necessary to explicitly apply custom classes anymore to the parent element of a component since it's done automatically in vue 3

fofi (approved):
LGTM

Navas23 (approved):

PR #155, opened 2025-09-15
Vue 3 migration: Use vitepress

dletran:
Vue 3 migration:
This is the 3rd of 3 PRs:

1. Migration prerequisites
2. Lib update
3. **Doc tool migration**

In this one we're migrating the documentation tool since Vuepress is not compatible with Vue 3. We're migrating to Vitepress since we can make use of most of the things we already had for Vuepress with a few changes:
- Vitepress config
  - Changed some imports syntax so they're compatible with Vite (browser env instead of node env) 
- Examples generation:
  - Import VueLive component to replace vue-live plugin not available in Vuepress
  - Changed the example template + added hygen helper function to adapt existing examples syntax to new VueLive component
- Others:
  - Created new badge component not existing in Vitepress


src/elements/GeoCalendar/GeoCalendarGrid/GeoCalendarDayGrid.vue
```diff
@@ -32,16 +32,7 @@
...
-import eachDayOfInterval from 'date-fns/eachDayOfInterval'
-import endOfWeek from 'date-fns/endOfWeek'
-import endOfMonth from 'date-fns/endOfMonth'
-import format from 'date-fns/format'
-import getDay from 'date-fns/getDay'
-import startOfMonth from 'date-fns/startOfMonth'
-import startOfWeek from 'date-fns/startOfWeek'
-import subDays from 'date-fns/subDays'
-import isValid from 'date-fns/isValid'
+import { addDays, eachDayOfInterval, endOfWeek, endOfMonth, format, getDay, startOfMonth, startOfWeek, subDays, isValid } from 'date-fns'
```
fofi, line 35: We used to had the imports like this to avoid including on the bundle the entire 'date-fns' lib, jsut the methods we actually use.
Has this changed?

src/utils/mockFontAwesomeProIcons.js
```diff
@@ -1,12 +1,5 @@
...
-  try {
-    return require('@fortawesome/free-solid-svg-icons')
-  } catch (error) {
-    return null
-  }
-})()
-
-const _ = require('lodash')
+import * as fas from '@fortawesome/free-solid-svg-icons'
+import _ from 'lodash'
```
fofi, line 2: Why we have to import lodash everywhere now?

.gitignore
```diff
@@ -18,13 +18,22 @@ yarn-error.log*
...
-.vuepress/components/constants
+.vitepress/cache/
+.vitepress/docs/components/
+.vitepress/docs/tokens/
+.vitepress/docs/guides/
+.vitepress/docs/design-tokens.md
+.vitepress/docs/elements.md
+.vitepress/docs/changelog.md
+.vitepress/docs/index.md
+!.vitepress/docs/public/
```
Navas23: Put it first and comment it pls

.hygen.js
```diff
@@ -191,7 +202,7 @@ module.exports = {
 
       const json = _.map(documentation.slots, function (slotMetadata, slotName) {
```
Navas23: We are not using slotName anymore, right?

.hygen.js
```diff
@@ -205,6 +216,102 @@ module.exports = {
...
+    /**
+     * Generates example files and documentation with ?raw imports
+     * @param {string} exampleMarkdownPath - Path to the examples markdown file
+     * @param {string} componentName - Component name for folder structure
+     * @param {string} existingContent - Existing content of the target file
+     * @returns {string} - Generated documentation with example imports
+     */
+    generateExamplesFromFile (exampleMarkdownPath, componentName, existingContent = '') {
+      const content = fs.readFileSync(exampleMarkdownPath).toString()
+      const exampleNamePrefix = exampleMarkdownPath.split('/').pop().split('.').slice(0, -1).join('.')
```
Navas23: We should split this so it's safer and easier to read

.vitepress/theme/components/Badge.vue
```diff
@@ -0,0 +1,36 @@
...
+  props: {
+    text: {
+      type: String,
+      required: true
+    }
+  },
+  computed: {
+    badgeColor() {
+      // Replicating colors used in AllElementsListItem.vue
+      // TODO: use DS tokens instead
```
Navas23, line 25: If we are not doing it in this task, let put a task id here and note it down in the jira task so we don't forget about it

.vitepress/theme/components/ComponentDocumentationSlotsRow.vue
```diff
@@ -27,7 +27,7 @@ export default {
   },
   computed: {
     markdownDescription () {
-      return unescapeJSONString(this.description)
+      return unescapeJSONString(this.description || '')
```
Navas23: Wdyt if we make this method to support receiving a nil value

.vitepress/theme/components/LiveCode.vue
```diff
@@ -0,0 +1,144 @@
...
+        :class="{ 'live-code__toggle--active': showCode }"
+      >
+        {{ showCode ? 'Hide Code' : 'Show Code' }}
+      </button>
+    </div>
+  </div>
+</div>
+</template>
+
+<script setup>
```
Navas23: Are we forced to use vue3 sintax?

.vitepress/theme/index.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    })
+    
+    // Register tokens mixin
+    app.mixin(tokensMixin)
+    
+    // Register external components
+    app.component('VueLive', VueLive)
+    app.component('FontAwesomeIcon', FontAwesomeIcon)
+
```
Navas23: Remove empty lines

_templates/vitepress/component/documentation.ejs.t
```diff
@@ -0,0 +1,65 @@
...
+to: .vitepress/docs/components/<%= h.withoutExtension(path) %>.md
+---
+
+<% const componentName = h.path.dirname(path).split('/').pop() %>
+
+# <%= h.inflection.classify(componentName) %> <badge text="Ready" type="tip" /> <badge text="2.0.0+" />
+
+<%= h.getComponentDescription(jsonDocumentation) %>
+
+<% if (h.getComponentPropertiesJSON(jsonDocumentation) != '[]' ) { %>
```
Navas23, line 11: Just to understand, why do we compare against '[]' ?

package.json
```diff
@@ -164,12 +165,14 @@
...
+    "vitepress": "^1.6.4",
     "vue": "^3.4.0",
     "vue-directive-tooltip": "~1.6.3",
-    "vue-docgen-api": "^3.19.5",
+    "vue-docgen-api": "^4.79.2",
     "vue-loader": "^17.0.0",
     "vue-style-loader": "^4.1.2",
     "vuedraggable": "2.24.3",
+    "vue-live": "2.5.4",
     "vuepress": "^1.9.10",
```
Navas23: Don't we remove vuepress?

src/utils/mockFontAwesomeProIcons.js
dletran: we already had it here before, I just changed it from a require to import to be compatible with Vite

maybe you noticed I added it in some examples files, this is because the LiveCode component needs to be send any external libs through a prop to consider them safe to execute in the browser and it can't use the global lodash
also in vitepress components because I didn't configure it globally for vitepress components

src/elements/GeoCalendar/GeoCalendarGrid/GeoCalendarDayGrid.vue
dletran: I changed it due to Vite build failing when importing it like that, from what I checked [date-fns v2.7](https://date-fns.org/v2.7.0/docs/ECMAScript-Modules) (the one we use) is compatible with tree-shaking so it shouldn't be an issue

_templates/vitepress/component/documentation.ejs.t
dletran: I just copied this from the vuepress documentation template we had, my guess is that hygen compares empty arrays this way easier

.vitepress/theme/components/Badge.vue
dletran: I think it's not really something we'll do in the improvements task
it's not really mandatory since this is for the documentation site, just a comment in case in the future we change this and we can reuse tokens

.vitepress/theme/components/LiveCode.vue
dletran: not really, I changed it to the options API we're more used to

fofi (approved):
LGTM

Navas23 (approved):

PR #156, opened 2025-09-16
Update button label styles

Navas23:
The vertical alignment of button label was done by hand with padding. Now we use the flex one

### DIFF
--- a/src/styles/elements/GeoButton/_geo-button.scss
+++ b/src/styles/elements/GeoButton/_geo-button.scss
@@ -167,7 +167,7 @@
   user-select: none;
 
   .geo-button__label {
-    padding: 0 .83em;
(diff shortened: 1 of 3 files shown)


src/styles/elements/GeoButton/_geo-button.scss
```diff
@@ -162,13 +162,12 @@
...
   height: 30px;
   letter-spacing: 1px;
-  line-height: 1.25em;
   position: relative;
   text-transform: uppercase;
   user-select: none;
 
   .geo-button__label {
-    padding: .58em .83em .67em .83em;
+    padding: 0 .83em;
```
dletran: could we use a spacing token here instead of the .83em?

dletran (approved):

